PR #1, opened 2022-04-06
Add github actions back

dottorblaster:
Just reenabling github actions, especially the CI.


nelsonkopliku (approved):

dottorblaster:
Note to self:

- bring back `gh_release_to_obs_changeset.py
- bring back env variables we had in the old project

PR #4, opened 2022-04-07
Instruct the specfile to only create the RPM package for the agent binary

dottorblaster:
As above, we are just shipping the binary inside the RPM now, therefore we need less and less stuff. 

No node_modules needed anymore.

### DIFF
--- a/packaging/suse/trento.spec
+++ b/packaging/suse/trento.spec
@@ -26,15 +26,10 @@ Group:          System/Monitoring
 URL:            https://github.com/trento-project/trento
 Source:         %{name}-%{version}.tar.gz
 Source1:        vendor.tar.gz
-Source2:        node_modules.spec.inc
-Source3:        package.json
-%include        %_sourcedir/node_modules.spec.inc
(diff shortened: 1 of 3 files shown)


packaging/suse/trento.spec
```diff
@@ -26,17 +26,14 @@ Group:          System/Monitoring
...
 ExclusiveArch:  aarch64 x86_64 ppc64le s390x
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  golang-packaging
 BuildRequires:  golang(API) = 1.16
-BuildRequires:  npm
-BuildRequires:  local-npm-registry
 Requires:       golang-github-prometheus-node_exporter
 Provides:       %{name} = %{version}-%{release}
+Provides:       trento = %{version}-%{release}
+Obsoletes:      trento < %{version}-%{release}
```
stefanotorresi, line 36: We'll probably have to also include `trento-premium` here in the future, although I'm not sure yet. Just noting this down for our future selves.

stefanotorresi (approved):
🔝

PR #21, opened 2022-04-19
Publish csp information of a discovered pacemaker cluster

nelsonkopliku:
This PR determines the CSP of a cluster.

Extracts the information the same way we do for the cloud discovery of a host.

Clearly the information published by the Designated Controller will be the truth.

Question: what if no CSP was provided? No big issue from agent side, as just an empty string will be published.
What about the checks engine?
Gonna check on this later on the control plane

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/trento-project/agent/internal/cloud"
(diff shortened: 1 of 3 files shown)


fabriziosestito (approved):

fabriziosestito:
@arbulu89 
> Question: what if no CSP was provided? No big issue from agent side, as just an empty string will be published.
What about the checks engine?

Interesting question, how does the runner deal with no provider provided?

PR #23, opened 2022-04-19
Flat map sap systems payload lists

arbulu89:
Flat map SAP systems payload some lists.
- `Instances` list
- `SAP control Instances` list
- `SAP control processes` list
- `SAP properties` list
The key of those maps was redundant in fact.

Using a map was causing issues, as for example, the instances names (used as key) might not be unique, so the data was being overidden. This scenario is present when 2 instances of the same system are installed in the same host (we already have some user with this scenario).

I think this format of payload has more advantages, and matches better with the current backend system (as we don't have anymore big limitations to filter by fields, as we had at the beginning when consul was used as data storage).

PD: Backend changes coming soon!


nelsonkopliku (approved):
LGTM! Not really thorough review as I am focused on cheks, but looks good.

PR #33, opened 2022-05-05
Fix URL in the package spec

mpagot:
Let the URL description field to point to the new trento-project/agent dedicated repo


nelsonkopliku (approved):
Whoa, good catch!
Thank you very much :rocket:

dottorblaster (approved):
Great!

PR #39, opened 2022-06-02
Add a debug trace to know why the cluster data is not built

arbulu89:
For debugging purposes, add a new logging entry to print why the a cluster is not discovered in the host.
In the last case, it was not discovered because the `authkey` file was not there.

Adding it as an debug, because not having cluster might be normal in many nodes, so I don't want to spam errors if the logging is not set as debug.


dottorblaster (approved):
LGTM

PR #42, opened 2022-06-09
Discover aws cloud data

arbulu89:
A lot of things for a small feature hehe

This PR aims to send AWS metadata payload to the server. The metadata is obtained using the local metadata endpoint `http://169.254.169.254/latest/meta-data`

It usage is pretty uncomfortable. The first output gives you a list (as a directory) of new options, which you need to request again, to get the final values (this becomes recursive).
More information in:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-categories.html

The PR details:

- Extract the `client` variable to a new file to make it accessible in both providers files (it would be accessible either way, but this way it looks more generic)
- `aws.go` file implement the data discovery. Basically it requests the metadata endpoint recursively until getting all the entries. By now, we don't use all of them, but it was cleaner to make this code rather than calling individual requests (and memorize them..)
- The data is loaded in a `map[string]interface{}`, so in order to use it easier, it is decoded into the `AwsMetadata` struct. To do so, we need to first convert the map object to a json string, and later on unmarshall it. Yes, it doesn't look efficient at all, but I didn't find any other way. We could use `mapstructure` for this as we did in the past, but we already removed that dependency and I didn't want to add it back (this makes me become aware that we have tons of `mapstructure` annotations all around the code)
- After that, the loaded raw data is transformed to a dto, and sent it to the server. The dto makes the usage of the payload easy in the server side (it would be easy with the raw data as well, as elixir is super nice to do this with the pattern matching, but the code looks cleaner with the dto)
- Finally, a lot of testing... Most of the included files are fixtures for testing purposes.

At the end the payload looks like this:

```
{"agent_id":"12345","discovery_type":"cloud_discovery","payload":{"Provider":"aws","Metadata":{"account_id":"12345","ami_id":"ami-12345","availability_zone":"eu-west-1a","data_disk_number":1,"instance_id":"i-1234","instance_type":"t3.micro","region":"eu-west-1","vpc_id":"vpc-1234"}}} 
```


nelsonkopliku (approved):
Alright, LGTM.
Does the web need to be updated in order to support this new type of metadata?

internal/cloud/aws.go
```diff
@@ -0,0 +1,126 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+)
+
+const (
+	awsMetadataUrl      = "http://169.254.169.254/latest/"
```
dottorblaster, line 18: Isn't there any domain which we can point our requests too? Only IPs?

dottorblaster (approved):
LGTM overall but I have a question

internal/cloud/aws.go
dottorblaster: From the documentation looks like there isn't. Blimey :-(

arbulu89: No, there is not. It looks like something standard in cloud providers, as azure does it in the same way

PR #47, opened 2022-06-21
Get the agent ID in the main agent code package

arbulu89:
In order to start introducing fact gathering in the Agent side, we need to know the Agent ID beforehand, so we can do more operations (like start consuming from rabbitmq in a queue with the AgentID name).

Now, the Agent ID is attached to the discovery publishing, and this limits its usage. 
The PR aims to make the Agent ID more broadly available in the agent code.

### DIFF
--- a/internal/agent.go
+++ b/internal/agent.go
@@ -3,17 +3,25 @@ package internal
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


fabriziosestito (approved):

CDimonaco (approved):
LGTM

PR #53, opened 2022-07-11
Change trento-premium to be obsolete in the spec

rtorrero:
This PR simply changes the relationship in the spec file with the old `trento-premium` package so that it's now marked as Obsolete. This should reduce the possibility of users installing the older package without noticing that it has been obsoleted.

### DIFF
--- a/packaging/suse/trento-agent.spec
+++ b/packaging/suse/trento-agent.spec
@@ -33,7 +33,9 @@ BuildRequires:  golang(API) = 1.16
 Requires:       golang-github-prometheus-node_exporter
 Provides:       %{name} = %{version}-%{release}
 Provides:       trento = %{version}-%{release}
+Provides:       trento-premium = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):

ajaeger (commented):
AFAIR this is not enough, see the linked documents.

packaging/suse/trento-agent.spec
```diff
@@ -33,7 +33,9 @@ BuildRequires:  golang(API) = 1.16
 Requires:       golang-github-prometheus-node_exporter
 Provides:       %{name} = %{version}-%{release}
 Provides:       trento = %{version}-%{release}
+Provides:       trento-premium = %{version}-%{release}
 Obsoletes:      trento < %{version}-%{release}
```
rtorrero, line 37: Now that I think about this, I suspect we should have probably not used the macro here to fix the (deprecated) provided version to an specific version.

PR #45, opened 2022-06-15
Spike checks engine

arbulu89:
Here the work I have started doing around the facts gathering logic we want to include in the agent side.

It is just a spike to see if this kind of code is what we are looking for.
It doesn't implement any kind of interaction with other components. Just dummy code.

In essence:
- It starts a the checks engine (maybe I should rename it to `factgatheringengine` ...)
- It has some placeholder function to subscribe to some virtual service and start receiving fact gathering requests. AKA: the `Listen` method
- It implements a dummy fact gathering to read the sbd config file output. It doesn't really do any side checking, like even if the file exists
- Build and send some basic struct with the results.
- A hidden `checkengine` flag for the command line to enable this feature (some sort of feature flag)

In this PoC, I have designed 2 basic struct, that are obviously open to change once we have the DSL defined.
1.
Fact gathering execution request. A list of facts that we want to gather. Something like:
```
# sbd_config is a hardcoded key to identify which fact gathering to do
# keys is a list with the keys from the list to return
[{"name": "sbd_config", "keys": ["SBD_DEVICE", "SBD_TIMEOUT_ACTION"]}]
```

2.
Fact results:
```
# name is the fact gathering identifier
# key is the obtained key
# value is the value of that key
[{"name":"sbd_config","key":"SBD_DEVICE","value":null},{"name":"sbd_config","key":"SBD_TIMEOUT_ACTION","value":"flush,reboot"}]
```

Feel free to have a look.
We can even throw this code away hehe

PR #54, opened 2022-07-11
Facts engine

arbulu89:
Inherited from https://github.com/trento-project/agent/pull/45, mostly polished and tested code.

It includes:
- corosync.conf fact gathering
- Core facts engine
- RabbitMQ adapter
- command line flags
- Docker-compose

### DIFF
--- a/internal/agent.go
+++ b/internal/agent.go
@@ -80,13 +80,13 @@ func getAgentID() (string, error) {
 
 // Start the Agent. This will start the discovery ticker and the heartbeat ticker
 func (a *Agent) Start(ctx context.Context) error {
-	g, ctx := errgroup.WithContext(ctx)
+	g, groupCtx := errgroup.WithContext(ctx)
 
 	for _, d := range a.discoveries {
 		dLoop := d
 		g.Go(func() error {
 			log.Infof("Starting %s loop...", dLoop.GetId())
-			a.startDiscoverTicker(ctx, dLoop)
+			a.startDiscoverTicker(groupCtx, dLoop)
 			log.Infof("%s discover loop stopped.", dLoop.GetId())
 			return nil
 		})
@@ -94,7 +94,7 @@ func (a *Agent) Start(ctx context.Context) error {
 
 	g.Go(func() error {
 		log.Info("Starting heartbeat loop...")
-		a.startHeartbeatTicker(ctx)
+		a.startHeartbeatTicker(groupCtx)
 		log.Info("heartbeat loop stopped.")
 		return nil
 	})
@@ -107,7 +107,7 @@ func (a *Agent) Start(ctx context.Context) error {
 				return err
 			}
 
-			if err := c.Listen(ctx); err != nil {
+			if err := c.Listen(groupCtx); err != nil {
 				return err
 			}
 
--- a/internal/agent.go
+++ b/internal/agent.go
@@ -29,8 +29,6 @@ type Agent struct {
 	config          *Config
 	collectorClient collector.Client
 	discoveries     []discovery.Discovery
-	ctx             context.Context
-	ctxCancel       context.CancelFunc
 }
 
 type Config struct {
@@ -59,14 +57,10 @@ func NewAgent(config *Config) (*Agent, error) {
 		discovery.NewHostDiscovery(collectorClient, *config.DiscoveriesConfig),
 	}
 
-	ctx, ctxCancel := context.WithCancel(context.Background())
-
 	agent := &Agent{
 		agentID:         agentID,
 		config:          config,
 		collectorClient: collectorClient,
-		ctx:             ctx,
-		ctxCancel:       ctxCancel,
 		discoveries:     discoveries,
 	}
 	return agent, nil
@@ -85,7 +79,7 @@ func getAgentID() (string, error) {
 }
 
 // Start the Agent. This will start the discovery ticker and the heartbeat ticker
-func (a *Agent) Start() error {
+func (a *Agent) Start(ctx context.Context) error {
 	var wg sync.WaitGroup
 
 	for _, d := range a.discoveries {
@@ -93,7 +87,7 @@ func (a *Agent) Start() error {
 		go func(wg *sync.WaitGroup, d discovery.Discovery) {
 			log.Infof("Starting %s loop...", d.GetId())
 			defer wg.Done()
-			a.startDiscoverTicker(d)
+			a.startDiscoverTicker(ctx, d)
 			log.Infof("%s discover loop stopped.", d.GetId())
 		}(&wg, d)
 	}
@@ -102,7 +96,7 @@ func (a *Agent) Start() error {
 	go func(wg *sync.WaitGroup) {
 		log.Info("Starting heartbeat loop...")
 		defer wg.Done()
-		a.startHeartbeatTicker()
+		a.startHeartbeatTicker(ctx)
 		log.Info("heartbeat loop stopped.")
 	}(&wg)
 
@@ -113,7 +107,7 @@ func (a *Agent) Start() error {
 			log.Info("Starting fact gathering service...")
 			defer wg.Done()
 			c.Subscribe()
-			c.Listen(a.ctx)
+			c.Listen(ctx)
 			log.Info("fact gathering stopped.")
 		}(&wg)
 	}
@@ -123,12 +117,12 @@ func (a *Agent) Start() error {
 	return nil
 }
 
-func (a *Agent) Stop() {
-	a.ctxCancel()
+func (a *Agent) Stop(ctxCancel context.CancelFunc) {
+	ctxCancel()
 }
 
 // Start a Ticker loop that will iterate over the hardcoded list of Discovery backends and execute them.
-func (a *Agent) startDiscoverTicker(d discovery.Discovery) {
+func (a *Agent) startDiscoverTicker(ctx context.Context, d discovery.Discovery) {
 
 	tick := func() {
 		result, err := d.Discover()
@@ -138,11 +132,11 @@ func (a *Agent) startDiscoverTicker(d discovery.Discovery) {
 		}
 		log.Infof("%s discovery tick output: %s", d.GetId(), result)
 	}
-	Repeat(d.GetId(), tick, time.Duration(d.GetInterval()), a.ctx)
+	Repeat(d.GetId(), tick, time.Duration(d.GetInterval()), ctx)
 
 }
 
-func (a *Agent) startHeartbeatTicker() {
+func (a *Agent) startHeartbeatTicker(ctx context.Context) {
 	tick := func() {
 		err := a.collectorClient.Heartbeat()
 		if err != nil {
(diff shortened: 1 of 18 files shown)


internal/agent.go
```diff
@@ -58,6 +62,7 @@ func NewAgent(config *Config) (*Agent, error) {
 	ctx, ctxCancel := context.WithCancel(context.Background())
```
CDimonaco, line 62: Could we think of adding the context as parameter to the constructor function? Maybe we can tweak and wire all the context in the entrypoint and pass a single context when we need it during the execution

internal/agent.go
```diff
@@ -101,6 +106,18 @@ func (a *Agent) Start() error {
 		log.Info("heartbeat loop stopped.")
 	}(&wg)
 
+	if a.config.FactsEngineEnabled {
```
CDimonaco, line 109: This goroutines are somehow wired together? In terms of errors propagation? If yes we can think of using a errgroup with the same context, in order to coordinate the failure/context injection. The code should be the same, I think that we can get rid of the waitgroup also, the errgroup has this similar waitgroup implementation inside.

internal/factsengine/adapters/rabbitmq.go
```diff
@@ -0,0 +1,79 @@
...
+	gatherFactsExchanage string = "gather_facts"
+	factsExchanage       string = "facts"
+)
+
+type rabbitMQAdapter struct {
+	consumer  rabbitmq.Consumer
+	publisher *rabbitmq.Publisher
+}
+
+func NewRabbitMQAdapter(factsEngineService string) *rabbitMQAdapter {
```
CDimonaco, line 18: Could we expose the rabbitMQAdapter struct?, generally it's good pratice to expose the struct if we have some function that returns that struct, this does not means that the field should be exposed but only the struct itself. This issue is very common, and there are some linting rules that can be used in order to avoid that. Like the default rules of golangci-lint

internal/factsengine/adapters/rabbitmq.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return &rabbitMQAdapter{
+		consumer:  consumer,
+		publisher: publisher,
+	}
+}
+
+func (r *rabbitMQAdapter) Unsubscribe() error {
+	r.consumer.Close()
```
CDimonaco, line 44: Could we check the errors returned by Close? We can return it and spot some problems

internal/factsengine/factsengine.go
```diff
@@ -0,0 +1,174 @@
...
+	//RabbitMQ adapter exists only by now
+	c.factsServiceAdapter = adapters.NewRabbitMQAdapter(c.factsEngineService)
+	log.Infof("Subscription to the facts engine by agent %s in %s done", c.agentID, c.factsEngineService)
+
+	return nil
+}
+
+func (c *factsEngine) Unsubscribe() error {
+	log.Infof("Unsubscribing agent %s from the facts engine service", c.agentID)
+	c.factsServiceAdapter.Unsubscribe()
```
CDimonaco, line 42: we should check the error here, and even in other defer clause similar to that, with a function and just log the error

internal/factsengine/factsengine.go
```diff
@@ -0,0 +1,174 @@
...
+	log.Infof("Unsubscribed properly")
+
+	return nil
+}
+
+func (c *factsEngine) Listen(ctx context.Context) {
+	log.Infof("Listening for facts gathering events...")
+	defer c.Unsubscribe()
```
CDimonaco, line 50: like

 ```go
  defer func(){
    err := c.Unsubscribe()
    // log me and go ahead!
  }()
```

File: internal/factsengine/factsengine.go [hunk 7]
```diff
@@ -0,0 +1,174 @@
...
+
+	if err := c.publishFacts(gatheredFacts); err != nil {
+		log.Errorf("Error publishing facts: %s", err)
+		return err
+	}
+
+	return nil
+}
+
+func gatherFacts(groupedFactsRequest *gatherers.GroupedFactsRequest, factGatherers map[string]gatherers.FactGatherer) (*gatherers.FactsResult, error) {
```
CDimonaco, line 81: maybe we can use a errgroup here to coordinate all the goroutines and do one error handling and one waiting

internal/factsengine/factsengine.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	log.Infof("Gathered facts response built properly")
+
+	return jsonFacts, nil
+}
+
+func prettyString(str []byte) string {
+	var prettyJSON bytes.Buffer
+	json.Indent(&prettyJSON, str, "", "  ")
```
CDimonaco, line 154: we should check the error here from Indent

internal/agent.go
arbulu89: Done!
Thank you for the tip

internal/factsengine/adapters/rabbitmq.go
arbulu89: Done!

internal/factsengine/adapters/rabbitmq.go
arbulu89: Done!

internal/factsengine/factsengine.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+type factsEngine struct {
+	agentID             string
+	factsEngineService  string
+	factGatherers       map[string]gatherers.FactGatherer
+	factsServiceAdapter adapters.Adapter
+}
+
+func NewFactsEngine(agentID, factsEngineService string) *factsEngine {
```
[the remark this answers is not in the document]
arbulu89: Done!

internal/factsengine/factsengine.go
arbulu89: Done!

internal/factsengine/factsengine.go
arbulu89: Done

internal/factsengine/gatherers/corosyncconf.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func corosyncConfToMap(lines []string) (map[string]interface{}, error) {
+	var corosyncMap = make(map[string]interface{})
+	var sectionStartPattern = `^\s*(\w+)\s*{.*`
+	var sectionEndPattern = `^\s*}.*`
+	var valuePattern = `^\s*(\w+)\s*:\s*(\S+).*`
+	var sections int = 0
+
+	sectionStartPatternCompiled := regexp.MustCompile(sectionStartPattern)
```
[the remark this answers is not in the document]
arbulu89: Done!
I just moved the compilation to the top level of the package. I don't know if you meant something else as well with this `use it as reference`, but I couldn't figure it out

internal/factsengine/gatherers/facts.go
```diff
@@ -0,0 +1,27 @@
+package gatherers
+
+type FactsResult struct {
+	ExecutionID string  `json:"execution_id"`
+	Facts       []*Fact `json:"facts"`
```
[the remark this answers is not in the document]
arbulu89: Removed pointers from the slices

arbulu89:
Hey @CDimonaco ,
Changes done! Could you have a new review?
As the next step, we could install the golang linter as you propose and set it up in the CI, so we catch all the issues here.
My proposal is that you do it @CDimonaco XD, as you have much more experience than anyone else in this field

internal/agent.go
```diff
@@ -80,38 +79,52 @@ func getAgentID() (string, error) {
 }
 
 // Start the Agent. This will start the discovery ticker and the heartbeat ticker
-func (a *Agent) Start() error {
-	var wg sync.WaitGroup
+func (a *Agent) Start(ctx context.Context) error {
+	g, ctx := errgroup.WithContext(ctx)
```
CDimonaco, line 83: shadowing, and reassigning the ctx variable from parameter can lead to some smell, maybe we can rename the ctx of the errogroup to something like groupCtx, and use it in the flow like you did below, so we can distinguish the two contexts

internal/agent.go
```diff
@@ -80,38 +79,52 @@ func getAgentID() (string, error) {
...
+				return err
+			}
+
+			log.Info("fact gathering stopped.")
+			return nil
+		})
+	}
 
-	return nil
+	return g.Wait()
```
CDimonaco, line 119: If we want to be more clear, we can wait for the errgroup, and add an if to check the error, and errors.Is if the error is context canceled, so we can log and maybe later distinguish when the error is context canceled (most of the times is a legit operation done by us) and when the error is unexpected

example 

```go
err = g.Wait()
if err != nil {
 if errors.Is(err, context.Canceled){
  // our fault? log this specific case
}
// anyway return the err
return err
}
// errgroup done without error
return nil
```

internal/factsengine/adapters/rabbitmq.go
```diff
@@ -0,0 +1,79 @@
...
+	r.publisher.Close()
+
+	return nil
+}
+
+func (r *rabbitMQAdapter) Listen(agentID string, handle func([]byte) error) error {
```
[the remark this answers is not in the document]
CDimonaco: perfect

internal/factsengine/gatherers/corosyncconf.go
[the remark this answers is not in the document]
CDimonaco: I was meaning move to the package level as variables, sorry for the misunderstanding, looks great

internal/factsengine/factsengine.go
```diff
@@ -0,0 +1,201 @@
...
+		if !exists {
+			log.Errorf("Fact gatherer %s does not exist", gathererType)
+			continue
+		}
+
+		// Execute the fact gathering asynchronously and in parallel
+		g.Go(func() error {
+			newFacts, err := gatherer.Gather(factsRequest)
+			if err == nil {
+				factsResults.Facts = append(factsResults.Facts, newFacts[:]...)
```
CDimonaco, line 120: appending an array in a goroutine like that can lead to some race conditions or unexpected behavior, we should use a channel that the goroutine uses to send the new facts, and have a channel loop in the gather facts function that collect this result and append in a single place, in a single routine to the facts inside the factsResult struct.

At first glance this can work, but there is the risk, there are multiple goroutine involved, and they can lead to a race

internal/agent.go
CDimonaco: Perfect, we can track it and we can deal with it later :D

internal/factsengine/factsengine.go
arbulu89: Good point!
Changed to use a buffered channel

internal/agent.go
arbulu89: Done

internal/factsengine/factsengine.go
CDimonaco: I don't know If I am clear about that, but we need another goroutine and the waitgroup, the other goroutine will defer the channel closing when all the routines sent the payload on channel.

CDimonaco (approved):
LGTM

PR #55, opened 2022-07-12
Identify SAP diagnostics agent

arbulu89:
Cherry-pick missing commit from legacy trento: https://github.com/trento-project/trento/commit/3467e77575395637a58c22606fad409bb1c15c78

I have done some additional changes to adapt the code, as we already were on some conflicts. But the code in essence is the same, so a deep review shouldn't be needed.

After this, some small changes in the web side are needed too.

### DIFF
--- a/internal/sapsystem/sapsystem.go
+++ b/internal/sapsystem/sapsystem.go
@@ -23,6 +23,7 @@ const (
 	Unknown = iota
 	Database
 	Application
+	DiagnosticsAgent
 )
 
 const (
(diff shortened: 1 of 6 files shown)


internal/sapsystem/sapsystem.go
```diff
@@ -485,6 +505,31 @@ func NewSAPInstance(w sapcontrol.WebService) (*SAPInstance, error) {
...
 }
 
+func detectType(sapControl *SAPControl) (int, error) {
+	sapLocalhost, err := sapControl.findProperty("SAPLOCALHOST")
+	if err != nil {
+		return Unknown, err
+	}
+
+	var instanceType int
+	for _, instance := range sapControl.Instances {
```
CDimonaco, line 515: As I understand here, we loop through the instances but instanceType will be always the last of instances that has sapLocalhost, because we don't break the loop, it's an expected behavior? Maybe if we are only interested in the "first one", or "at least one" of the instances is sapLocalhost we can return into the cases of the switch without looping.

Because if in the instances no one is the sapLocalhost instanceType will be the zero value of int so zero, maybe if this is expected we can initialize the variable not to int but to the appropriate "enum value" like "Application" "Database" etc..

nelsonkopliku (approved):
LGTM

PR #61, opened 2022-07-20
Linter configuration

CDimonaco:
Enable golangci-lint with a custom configuration.

Some code and tests refactoring is needed.
The test will be the guidance for the refactoring


cmd/root.go
```diff
@@ -33,12 +34,17 @@ that can help you deploy, provision and operate infrastructure for SAP Applicati
 	var cfgFile string
 	var logLevel string
 
-	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.trento.yaml)")
-	rootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "info", "then minimum severity (error, warn, info, debug) of logs to output")
+	rootCmd.PersistentFlags().
```
arbulu89, line 37: Nothing about linting, but here, we could replace `StringVar` by `String` simply, as we don't really use this variables in this functions, and we get them later on with `viper.GetString` (or similars).

cmd/start.go
```diff
@@ -21,36 +22,103 @@ func NewStartCmd() *cobra.Command {
...
+			&sshAddress,
+			"ssh-address",
+			"",
+			"The address to which the trento-agent should be reachable for ssh connection by the runner for check execution.",
+		)
 
-	startCmd.Flags().StringVar(&serverURL, "server-url", "http://localhost", "Trento server URL")
-	startCmd.Flags().StringVar(&apiKey, "api-key", "", "API key provided by trento control plane. Allows communication")
+	startCmd.Flags().
+		StringVar(
```
arbulu89, line 50: Same here. Better maybe to use `String`

internal/cloud/aws.go
```diff
@@ -54,10 +54,39 @@ type Placement struct {
 
 func NewAwsMetadata() (*AwsMetadata, error) {
 	var err error
-	awsMetadata := &AwsMetadata{}
+	awsMetadata := &AwsMetadata{
```
arbulu89, line 57: Dump question. I know that this is highligthed by the linter, but...
What is the advantage of declaring all the field?
It looks a lot a lot of boilerplate.
Does it improve the code somehow? Or simply readability?

internal/cluster/cib/parser.go
```diff
@@ -11,13 +11,13 @@ type Parser interface {
 	Parse() (Root, error)
 }
 
-type cibAdminParser struct {
+type AdminParser struct {
```
arbulu89, line 14: This changes makes the code a bit more confusing. This struct is used to store the `cibadmin` command output. That's why it was named as it was. Maybe we could have `CibAdminParser` and not lint it?

internal/cluster/crmmon/data.go
```diff
@@ -34,7 +34,7 @@ type Root struct {
 		Nodes []struct {
 			Name            string `xml:"name,attr"`
 			ResourceHistory []struct {
-				Name               string `xml:"id,attr"`
+				Name               string `xml:"id,attr" json:"Name"`
```
arbulu89, line 37: Some reason to add this `json` annotation? I don't think we use it anywhere

internal/cluster/sbd.go
```diff
@@ -38,7 +38,7 @@ type SBDDevice struct {
 
 type SBDDump struct {
 	Header          string `mapstructure:"header,omitempty"`
-	Uuid            string `mapstructure:"uuid,omitempty"`
+	UUID            string `mapstructure:"uuid,omitempty" json:"Uuid"`
```
arbulu89, line 41: I think we don't use `mapstructure` anymore, we could remove it.
About the `json` field, we should check if this can affect somehow in the server side code...
Most probably now, because we don't use them, but this `SnakeCase` things looks suspicious

internal/cluster/sbd.go
```diff
@@ -68,13 +68,19 @@ func FindMatches(pattern string, text []byte) map[string]interface{} {
 	r := regexp.MustCompile(pattern)
 	values := r.FindAllStringSubmatch(string(text), -1)
 	for _, match := range values {
-		key := strings.Replace(match[1], " ", "_", -1)
+		key := strings.ReplaceAll(match[1], " ", "_")
 		if _, ok := configMap[key]; ok {
 			switch configMap[key].(type) {
 			case string:
-				configMap[key] = []interface{}{configMap[key]}
+				{
```
arbulu89, line 75: Why this brackets? Is the linter doing it?
It looks uglier hehe

internal/hosts/metadata.go
```diff
@@ -2,11 +2,11 @@ package hosts
 
 type Metadata struct {
 	Cluster         string `mapstructure:"ha-cluster,omitempty"`
-	ClusterId       string `mapstructure:"ha-cluster-id,omitempty"`
+	ClusterID       string `mapstructure:"ha-cluster-id,omitempty" json:"ClusterId"`
```
arbulu89, line 5: Where do we use this `Metadata` struct?
I would bet that nowhere hehe

rtorrero:
Very good stuff @CDimonaco and @fabriziosestito, thanks!

arbulu89 (commented):
Thanks for this @CDimonaco !!
I have noted some comments and questions before approving.

internal/cluster/cib/parser.go
CDimonaco: The go way is to express something with the name of the package and the thing you are importing, you refer to that struct with cib.AdminParser, instead of cib.CibAdminParser, in this case the name of the package already express the cib domain so in order to adhere to the code standars you should omit Cib in the struct name.

Anyway, if there are some cases in which repeating something brings a lot more clarity, we can opt for that, but I don't think this is the case. Anyway I'm not an expert of this domain, so if you think that we should keep the prefix, we can leave it as is

internal/cluster/sbd.go
CDimonaco: Same as above, it's for mantaining the compatibility with the fixtures, the fixture expect Uuid in that casing, so I changed and linted in the go code, but I had to mantain the compatibility with json, it's strange and we should fix it, but it's another pr scope imho

internal/cluster/sbd.go
CDimonaco: Yes :D 😆

internal/cluster/cib/parser.go
arbulu89: In that case I would name it simply `Parser`. The `Admin` prefix doesn't make much sense if it is not coming with `Cib`.

internal/cloud/aws.go
arbulu89: Ok! Thank you

internal/cluster/crmmon/data.go
arbulu89: Ok. We could have some tech debt ticket in our backlog for this I guess

internal/cluster/sbd.go
arbulu89: Let's at least be careful on this changes, as this data might be consumed by the web server side

internal/hosts/metadata.go
arbulu89: Yes, yes, I was not complaining about the change. Just saying that this struct is most probably some legacy code we had when we used consul, but not it is not used anywhere, so removing it (together with the test) should be safe

cmd/root.go
CDimonaco: Done!

internal/cluster/cib/parser.go
CDimonaco: done

CDimonaco:
@arbulu89 Changes made, could you take another look before the merge?

cmd/start.go
CDimonaco: done!

arbulu89 (approved):
Perfect @CDimonaco !
Ready to merge from my side!
Having the `FIXME` as ticket in our backlog would be a plus XD

CDimonaco:
All done :D also the ticket

PR #63, opened 2022-07-26
Add plugins system

arbulu89:
Implement the fact gathering plugins system.

Most of the "How to use plugins" information added in the README file, so have a look there.

Besides that, the code simply adds the option to load plugins using [go-plugin](https://github.com/hashicorp/go-plugin) and a minimal RPC option. gRPC is not included yet, as it only enables other languages usage by now, and I think that unless we have some user asking it, we can skip it by now.

In any case, the code to use gRPC and even the native golang plugins system is more or less there (pending to add more PluginLoaders and relevant code).

Finally, some other additions:
- Update the rpm spec file to create the plugins folder
- Update the CI to test the plugins usage as more integration test.
- The go-plugin logging "defaulted" to use the same level provided to the trento agent command


CDimonaco (approved):
LGTM

nelsonkopliku (commented):
Nice stuff!

If I understand this correctly this is what happens:
- we have the agent binary with already some embedded gatherers, the ones we deliver
- the user can write his own gatherers (go files following a structure) 
- build those extra gatherers and place the output in a known location
- upon agent restart the user defined gatherers plugins will be loaded from the specified location

Am I getting this correctly? The Agent just needs to be restarted, correct?

arbulu89:
> Nice stuff!
> 
> If I understand this correctly this is what happens:
> 
>     * we have the agent binary with already some embedded gatherers, the ones we deliver
> 
>     * the user can write his own gatherers (go files following a structure)
> 
>     * build those extra gatherers and place the output in a known location
> 
>     * upon agent restart the user defined gatherers plugins will be loaded from the specified location
> 
> 
> Am I getting this correctly? The Agent just needs to be restarted, correct?

Yes, perfectly explained! XD

nelsonkopliku (approved):

PR #64, opened 2022-07-28
Remove the flavor field and add the installation source

arbulu89:
The `flavor` field is not used in the Agent since the web code was extracted, and there is not any "premium" capability.
For that reason it is removed.

And in order to have more information in our telemetry system, a new "installation source" metric is added, so we know how the agent was installed. The options are:

- `Community`: The agent was build after getting it from Github (whether downloading the code and building or getting the artficats)
- `Suse`: The agent is installed through a RPM package coming from OBS/IBS. There is not any distinctions between the source repository.

### DIFF
--- a/test/fixtures/discovery/host/expected_published_host_discovery.json
+++ b/test/fixtures/discovery/host/expected_published_host_discovery.json
@@ -13,6 +13,7 @@
         "cpu_count": 2,
         "socket_count": 1,
         "total_memory_mb": 4096,
-        "agent_version": "trento-agent-version"
+        "agent_version": "trento-agent-version",
+        "installation_source": "Community"
     }
 }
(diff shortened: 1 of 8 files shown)


test/fixtures/discovery/host/expected_published_host_discovery.json
```diff
@@ -13,6 +13,7 @@
         "cpu_count": 2,
         "socket_count": 1,
         "total_memory_mb": 4096,
-        "agent_version": "trento-agent-version"
```
CDimonaco, line 16: I'm just curious, this fixture are some api response? Or we use them in order to deep match a struct, like serialize the struct and deep match this fixtures in test?

arbulu89: This json is what the agent is supposed to send to the web side, to the `/api/collect` endpoint.

CDimonaco: Thanks for clarification! I'm still exploring the codebase 👀

CDimonaco:
LGTM

CDimonaco (approved):

PR #65, opened 2022-08-02
Add AgentID and CheckID fields to facts result

arbulu89:
Add AgentID and CheckID fields to facts request and result, on order to provide Wanda all the required data.

@CDimonaco I have created the new `CreateFactWithRequest` function in order to make the creation easier and more standardized. It is used in the `Gather` function of the `corosyncconf.go` file. I was thinking if we could enforce the usage of this method somehow in the `Gather` functions, so new gatherers don't miss all the fields. Do you know any trick we could have to enforce this? Maybe there isn't hehe, and we would need to validate later on all the fields are populated.

### DIFF
--- a/internal/factsengine/factsengine.go
+++ b/internal/factsengine/factsengine.go
@@ -113,7 +113,7 @@ func (c *FactsEngine) handleRequest(request []byte) error {
 		return err
 	}
 
-	gatheredFacts, err := gatherFacts(factsRequests, c.factGatherers)
+	gatheredFacts, err := gatherFacts(c.agentID, factsRequests, c.factGatherers)
 	if err != nil {
 		log.Errorf("Error gathering facts: %s", err)
 		return err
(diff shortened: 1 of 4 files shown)


CDimonaco:
We can't really enforce that, we need to validate the fields, of course we can do that in the new function, or we could use a separate Validate function, or we could use a builder pattern, and on the build method we perform all the validation.

My suggestion is to validate into the costructor function if we have few fields and we know upfront the validation problems that could occur on the creation

arbulu89:
> We can't really enforce that, we need to validate the fields, of course we can do that in the new function, or we could use a separate Validate function, or we could use a builder pattern, and on the build method we perform all the validation.
> 
> My suggestion is to validate into the costructor function if we have few fields and we know upfront the validation problems that could occur on the creation

Ok, thanks!
We will implement the validation them once we know exactly the "definitive" (for the 1st iteration) fields

nelsonkopliku (commented):
I came late to this party :sweat_smile: 
Anyway.

The addition of the missing fields looks good to me, if I remember correctly they're what the latest discussions made emerge.

Still, as per recent discussions on slack, we have a pending decision to be made which revolves around:
- event semantic for the messages exchanged by the components (ie integration events: something like: FactsGatheringRequested or FactsGatheringCompleted and similar)
- Error definition and representation
- schema for the messages (cloudevents|custom|others)

I guess we will iterate over.

PR #67, opened 2022-08-04
Corosync cmapctl gatherer

rtorrero:
This PR adds a gatherer to collect facts from the `corosync-cmapctl` binary and makes it available through our factsengine

Note that I had to move the utils to a separate package as I was incurring in cyclic dependencies when attempting it to use it in the internal package


internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,45 @@
+package gatherers
+
+import (
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/agent/internal/utils"
+)
+
+const (
+	CorosyncCmapCtlFactKey = "corosync-cmapctl"
```
arbulu89, line 9: In the package version gatherer I have used `_` as "splitter". We should agree on this hehe
`-` or `_`?

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,45 @@
...
+	return &CorosyncCmapctlGatherer{
+		executor: Executor{},
+	}
+}
+
+func (s *CorosyncCmapctlGatherer) Gather(factsRequests []FactRequest) ([]Fact, error) {
+	facts := []Fact{}
+	log.Infof("Starting corosync-cmapctl facts gathering process")
+
+	corosyncCmapctl, err := s.executor.Exec( // nolint:gosec
```
arbulu89, line 26: Why are we disabling linting here? I think we can remove it safely as the arguments are constants

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,45 @@
...
+		"corosync-cmapctl", "-b")
+	if err != nil {
+		return facts, err
+	}
+
+	corosyncCmapctlMap := utils.FindMatches(`(?m)^(\S*)\s\(\S*\)\s=\s(.*)$`, corosyncCmapctl)
+
+	for _, factReq := range factsRequests {
+		if value, ok := corosyncCmapctlMap[factReq.Argument]; ok {
+			facts = append(facts, Fact{
```
arbulu89, line 36: Let's use the new `NewFactWithRequest` function, so we create the facts with all the needed fields

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,45 @@
...
+	corosyncCmapctl, err := s.executor.Exec( // nolint:gosec
+		"corosync-cmapctl", "-b")
+	if err != nil {
+		return facts, err
+	}
+
+	corosyncCmapctlMap := utils.FindMatches(`(?m)^(\S*)\s\(\S*\)\s=\s(.*)$`, corosyncCmapctl)
+
+	for _, factReq := range factsRequests {
+		if value, ok := corosyncCmapctlMap[factReq.Argument]; ok {
```
arbulu89, line 35: What if the value is not found?
We should handle this scenario. Maybe by now, we can put a log, but at some point we need to send back a fact with this data (maybe a new define hack).

internal/factsengine/gatherers/corosynccmapctl_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+type CorosyncCmapctlTestSuite struct {
+	suite.Suite
+}
+
+func TestCorosyncCmapctlTestSuite(t *testing.T) {
+	suite.Run(t, new(CorosyncCmapctlTestSuite))
+}
+
```
arbulu89, line 20: Maybe we will need to have a test for the scenario that the wanted fact is not present (new unhandled scenario I comment in a previous comment)

arbulu89 (changes requested):
Nice @rtorrero ,
You will need to rebase from main to get the latest changes, which will break some tests I'm afraid hehe

Besides that, once this or the `package_version` gatherer PRs are merge, we will need to rebase back again, as they have duplicated code for the command executor.

Finally, this new gatherer must be enabled in the `factsengine` code, as corosyncconf is right now. Otherwise we cannot really use it.

I would try to compile the binary and run to see that it actually works:
```
./trento-agent facts gather --gatherer corosync-cmapctl --argument runtime.config.totem.consensus
```

internal/factsengine/gatherers/corosynccmapctl.go
rtorrero: it's not really splitting anything, but reflecting the name of the binary as-is in the same way `corosync.conf` is the fact key for the corosync configuration gatherer. I still understand the confusion and don't have a strong opinion on this, wdyt?

rtorrero:
Thanks for the review, indeed I was missing enabling the gatherer in the facts engine. Seems to work correctly now:

```
INFO[2022-08-04 11:24:01] Using config file: /etc/trento/agent.yaml    
INFO[2022-08-04 11:24:01] Starting corosync-cmapctl facts gathering process 
INFO[2022-08-04 11:24:01] Requested corosync-cmapctl facts gathered    
INFO[2022-08-04 11:24:01] Gathered fact for "corosync-cmapctl" with argument "runtime.config.totem.consensus": 
INFO[2022-08-04 11:24:01] {
  "name": "runtime.config.totem.consensus",
  "value": "36000",
  "check_id": ""
} 
```

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,47 @@
...
+			fact := NewFactWithRequest(factReq, fmt.Sprint(value))
+			facts = append(facts, fact)
+		} else {
+			log.Warnf("%s gatherer: requested fact %s not found", CorosyncCmapCtlFactKey, factReq.Argument)
```
arbulu89, line 41: I have used `log.Errorf` in other gatherers, but I guess it doesn't really matter that much.
Once we know how to populate the failed fact gatherings we will need to come back and finish the job hehe

arbulu89 (approved):
Perfect!
Green light from my side

PR #68, opened 2022-08-04
Implement crm_mon and cibadmin gatherers

arbulu89:
`crm_mon` and `cibadmin` gatherers.

In order to get the values, `xpath` is used. Given a xpath string, the value corresponding to this xpath is returned as path.
As both have xml as output, I have created this `xml` file which basically works with any generic xml output. 

Example:
```
 # ./trento-agent facts gather --gatherer crm_mon --argument "//resource[@resource_agent='stonith:external/sbd']/@role"
INFO[2022-08-04 15:04:02] Starting crmmon facts gathering process      
INFO[2022-08-04 15:04:02] Requested crmmon facts gathered              
INFO[2022-08-04 15:04:02] Gathered fact for "crm_mon" with argument "//resource[@resource_agent='stonith:external/sbd']/@role": 
INFO[2022-08-04 15:04:02] {
  "name": "//resource[@resource_agent='stonith:external/sbd']/@role",
  "value": "Started",
  "check_id": ""
} 
```


internal/factsengine/gatherers/cibadmin_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	p := &CibAdminGatherer{
+		executor: mockExecutor,
+	}
+
+	factRequests := []FactRequest{
+		{
+			Name:     "instance_number",
+			Gatherer: "cibadmin",
+			Argument: "//primitive[@type='SAPHana']/instance_attributes/nvpair[@name='InstanceNumber']/@value",
```
nelsonkopliku, line 43: `//primitive[@type='SAPHana']/instance_attributes/nvpair[@name='InstanceNumber']/@value`
`//primitive[@type='SAPHana']/instance_attributes/nvpair[@name='SID']/@value`

this is what will be there in the DSL, I am wondering if we could simplify that by expecting something different, something like just the `instanceNumber` or `SID`, since the rest is actually the same :thinking: 

Anyway, I believe it is enough as a starting point, yet, I'd reason about options to simplify.

internal/factsengine/gatherers/xml.go
```diff
@@ -0,0 +1,37 @@
...
+	for _, factReq := range factsRequests {
+		x, err := xmlpath.Compile(factReq.Argument)
+		if err != nil {
+			log.Errorf("Error compiling xpath: %s", factReq.Argument)
+			return nil, err
+		}
+
+		value, ok := x.String(root)
+		if !ok {
+			// TODO: Decide together with Wanda how to deal with errors. `err` field in the fact result?
```
nelsonkopliku, line 28: Good point, we need to make this agreement 

@trento-project/developers we need this agreement, shall we refine the thing? (there's an item here https://jira.suse.com/browse/CFSA-840)

internal/factsengine/gatherers/crmmon_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	p := &CrmMonGatherer{
+		executor: mockExecutor,
+	}
+
+	factRequests := []FactRequest{
+		{
+			Name:     "role",
+			Gatherer: "crm_mon",
+			Argument: "//resource[@resource_agent='stonith:external/sbd']/@role",
```
nelsonkopliku, line 43: Same thought as the previous one: would it make sense to simplify?

nelsonkopliku (approved):
Looks good :rocket: 

Two points for possible followups:
- error representation/handling agreement
- possible simplification of the xpath arguments

internal/factsengine/gatherers/cibadmin_test.go
nelsonkopliku: I see also how that would require also accepting `type`, which opens to what @rtorrero raised recently about the possibility to provide support for multiple `Arguments`.

arbulu89: Well, it turns out that these 2 examples are almost identical, but the xml output from cibadmin and crm_mon has a lot of paths. We couldn't really simply that much. At the end a xml output is a pretty generic thing, and the user would want to get any part of it.

Other examples:
```
//rcs_location[@node="node01"]/@rsc
//lrm_resource[@class="ocf]/lrm_rsc_op[@operation="monitor"]/@operation_key
```

At the end, the user would like to get any item from the xml. I think that trying to simply if would make the opposite, make it more difficult.
Finally, `xpath` is pretty well known standard (specially from the times where xml was more used), so I think keep in it as it is is the best solution

nelsonkopliku: I understand... let's just go further.

internal/factsengine/gatherers/xml.go
arbulu89: Yes, we need to refine it

PR #69, opened 2022-08-04
Implement systemd daemons state gatherer

arbulu89:
systemd states gatherer. I have decide to go with the golang `systemd` pacakge together with `dbus`. Maybe it is a bit of a overkill, as we could simply run the `systemctl is-active` command, but this options makes it more extensible, as we can easily gather other fields of the daemon state in the future.

Here the api of dbus for additional help: https://pkg.go.dev/github.com/coreos/go-systemd/v22/dbus

Example:
```
./trento-agent facts gather --gatherer systemd --argument corosync
INFO[2022-08-04 15:14:56] Starting systemd state facts gathering process 
INFO[2022-08-04 15:14:56] Requested systemd state facts gathered       
INFO[2022-08-04 15:14:56] Gathered fact for "systemd" with argument "corosync": 
INFO[2022-08-04 15:14:56] {
  "name": "corosync",
  "value": "inactive",
  "check_id": ""
} 

```

```
./trento-agent facts gather --gatherer systemd --argument k3s
INFO[2022-08-04 15:15:12] Starting systemd state facts gathering process 
INFO[2022-08-04 15:15:12] Requested systemd state facts gathered       
INFO[2022-08-04 15:15:12] Gathered fact for "systemd" with argument "k3s": 
INFO[2022-08-04 15:15:12] {
  "name": "k3s",
  "value": "active",
  "check_id": ""
} 

```


nelsonkopliku (approved):
Pretty clear :heavy_check_mark:

dottorblaster:
> Maybe it is a bit of a overkill

I think this is the correct approach instead, good shot.

dottorblaster (approved):
LGTM

PR #70, opened 2022-08-04
Implement the hacluster password verify gatherer

arbulu89:
This gatherer checks if the provided password is the correct one for the `hacluster` user.

As stated in the code:
```
By now, this gatherer works with a fixed user: hacluster
It checks if the provider password in the fact Argument field matches with the user password, being:
- true: the provided password is hacluster user password
- false: the provided password does not match the hacluster password
In order to make the checked user variable, the arguments field should be converted to some more
advanced struct to allow getting more data
```

By now, there is not any need to check the password for other users. The aim of this one is to basically check that the `hacluster` user password has been changes from the default one, so the cluster doesn't have such a big security hole

Find some more information about the implementation details and how passwords work in linux systems using the hashing way here:
https://linux-audit.com/password-security-with-linux-etc-shadow-file/

### DIFF
--- a/internal/factsengine/gatherers/verifypassword_test.go
+++ b/internal/factsengine/gatherers/verifypassword_test.go
@@ -1,0 +1,107 @@
+package gatherers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	mocks "github.com/trento-project/agent/internal/factsengine/gatherers/mocks"
+)
+
+type PasswordTestSuite struct {
+	suite.Suite
+	mockExecutor *mocks.CommandExecutor
+}
+
+func TestPasswordTestSuite(t *testing.T) {
+	suite.Run(t, new(PasswordTestSuite))
+}
+
+func (suite *PasswordTestSuite) SetupTest() {
+	suite.mockExecutor = new(mocks.CommandExecutor)
+}
+
+func (suite *PasswordTestSuite) TestPasswordGatherEqual() {
+	shadow := []byte("hacluster:$6$WFEgPAefduOyvLCN$MprO90En7b/" +
+		"cP8uJJpHzJ7ufTPjYuWoVF4s.3MUdOR9iwcO.6E3uCHX1waqypjey458NKGE9O7lnWpV/" +
+		"qd2tg1:19029::::::")
+
+	suite.mockExecutor.On("Exec", "getent", "shadow", "hacluster").Return(
+		shadow, nil)
+
+	verifyPasswordGatherer := NewVerifyPasswordGatherer(suite.mockExecutor)
+
+	factRequests := []FactRequest{
+		{
+			Name:     "hacluster",
+			Gatherer: "password",
+			Argument: "hacluster:linux",
+			CheckID:  "check1",
+		},
+	}
+
+	factResults, err := verifyPasswordGatherer.Gather(factRequests)
+
+	expectedResults := []Fact{
+		{
+			Name:    "hacluster",
+			Value:   true,
+			CheckID: "check1",
+		},
+	}
+
+	suite.NoError(err)
+	suite.ElementsMatch(expectedResults, factResults)
+}
+
+func (suite *PasswordTestSuite) TestPasswordGatherNotEqual() {
+	shadow := []byte("hacluster:$6$WFEgSAefduOyvLCN$MprO90En7b/" +
+		"cP8uJJpHzJ7ufTPjYuWoVF4s.3MUdOR9iwcO.6E3uCHX1waqypjey458NKGE9O7lnWpV" +
+		"/qd2tg1:19029::::::")
+
+	suite.mockExecutor.On("Exec", "getent", "shadow", "hacluster").Return(
+		shadow, nil)
+
+	verifyPasswordGatherer := NewVerifyPasswordGatherer(suite.mockExecutor)
+
+	factRequests := []FactRequest{
+		{
+			Name:     "hacluster",
+			Gatherer: "password",
+			Argument: "hacluster:linux",
+			CheckID:  "check1",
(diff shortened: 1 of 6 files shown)


internal/factsengine/gatherers/verifypassword_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	factRequests := []FactRequest{
+		{
+			Name:     "hacluster",
+			Gatherer: "password",
+			Argument: "linux",
+			CheckID:  "check1",
+		},
+	}
```
nelsonkopliku, line 39: Here's how the DSL would look like, right?
```
facts:
    - name: <whatever_name>
      gatherer: password
      argument: linux
```
I am wondering how I would read such a piece of DSL :thinking:

If I understand correctly what you say here
> The aim of this one is to basically check that the hacluster user password has been changes from the default one
>

The gathered fact is _hacluster user password has been changed from the default one_ and I understand that being the reason why the result is a boolean.
And the **default password** you mentioned is the provided argument ("linux" in the example)

Initially I was a bit confused by the checking nuance this gatherer has by password equality :sweat_smile:, yet I realize that the fact of interest is the knowledge that the password has changed (yes/no), not the current value, and that requires internally to do the checking.
I believe then that returned bool will be used in check evaluation itself.

So, just to make it closer to a real scenario, a user would write this piece of DSL
```
facts:
    - name: hacluster_password_changed
      gatherer: password
      argument: linux
```
It works, yet I still feel a bit confused.
Maybe `password` is too a generic name for this gatherer, considering we are assuming the user `hacluster`.
And, since we are hardcoding the user, what stops us from doing the same for the default password?
I mean...hardcode/config/whatever

Imagine we have this DSL
```
facts:
    - name: my_hacluster_pass_has_changed
      gatherer: hacluster_password_changed
```
By finding a more specific name for the gatherer in relation to what it is actually doing and encapsulating what is encapsulable, I believe we could have a clearer and more specific definition, also simpler to read imho :sweat_smile:

nelsonkopliku (commented):
Thanks Xabi, great stuff!
I shared with you some thoughts about the gatherer and related DSL.
Let's hear also from the rest of the team :rocket:

internal/factsengine/gatherers/verifypassword_test.go
arbulu89: The gatherer name is `verify_password`. 
So it would look like:
```
facts:
    - name: hacluster_password_changed
      gatherer: verify_password
      argument: linux
```

I was waiting to have this extra arguments option to have the option to get the user from the dsl itself. I don't know how useful it would be to check other users password. But I thought that having it this way would make it at least more usable.

Anyway, we can certainly hardcode the password and change the gatherer name.
But I still think that having it generic will be beneficial. Once we have the option to send the user, we could create more checks, and even our users could test if their own system users don't have default passwords, etc

nelsonkopliku: > I don't know how useful it would be to check other users password.
>
This is exactly it :smile: 
this sentence makes me think that the password changed check is very specific for a given user (`hacluster`), at least for now.

However, I understand the abstraction you're looking for, something like this I believe, which I find reasonable
```
facts:
    - name: hacluster_password_changed
      gatherer: verify_password_changed # maybe this clarifies more
      arguments: 
          user: hacluster
          password: linux
```
and I understand that the missing possibility to provide multiple arguments is a limitation.

There is another way forward.

I believe @rtorrero had a similar issue and the solution was providing only one argument that follows a known pattern.
Something like `username:password`

What do you think about that?

arbulu89: Ok, I will change to your suggestion.

rtorrero:
I'm a bit late to the party but I'm not entirely convinced that flexibility here is the key value to go for. I'd rather have the gatherer to check exclusively for a well-known pre-established password. e.g. user: hacluster, password:linux or whatever the known defaults are.

Sending user:password combinations to the gatherer is indeed more flexible and allows to reuse this gatherer for other things, but I'd seek for security over flexibility here.

As off for the code, if we still decide to go with this approach: LGTM

nelsonkopliku:
@rtorrero you mean something like this option?
```yaml
facts:
    - name: my_hacluster_pass_has_changed
      gatherer: hacluster_password_changed
```
that encapsulates username and related default password.

That was also my thought, let's hear also from the rest of the team.

nelsonkopliku:
I took the liberty to pick this up and:
- modified gatherer constructor to inject an executor
- updated tests

At this stage of current gatherer's implementation we are able to tell whether a `username:password` combination is correct (returns `true|false`)

The involved check here is [1.5.2 The `hacluster` user password has been changed from the default value `linux`](https://github.com/trento-project/runner/blob/main/runner/ansible/roles/checks/1.5.2/tasks/main.yml) and expressing that via DSL and this gatherer, would look like this
```
# 1.5.2 The `hacluster` user password has been changed from the default value `linux`
facts:
  - name: default_hacluster_password
    gatherer: verify_password
    argument: "hacluster:linux"
expectations:
  - name: hacluster_password_has_been_changed
    expect: default_hacluster_password == false
```
If this is  fine, let's go ahead, otherwise we'd need to change the gatherer's logic.

fabriziosestito (approved):

dottorblaster (approved):
LGTM

rtorrero:
> @rtorrero you mean something like this option?
> 
> ```yaml
> facts:
>     - name: my_hacluster_pass_has_changed
>       gatherer: hacluster_password_changed
> ```
> 
> that encapsulates username and related default password.
> 
> That was also my thought, let's hear also from the rest of the team.

Yes (sorry for the late reply)

PR #71, opened 2022-08-05
Implement saphostctrl gatherer

arbulu89:
saphostctrl gatherer. Simply returns the output of this commands.

Example: 
```
./trento-agent facts gather --gatherer saphostctrl --argument Ping
INFO[2022-08-05 09:12:00] Starting saphostctrl facts gathering process 
INFO[2022-08-05 09:12:00] Requested saphostctrl facts gathered         
INFO[2022-08-05 09:12:00] Gathered fact for "saphostctrl" with argument "Ping": 
INFO[2022-08-05 09:12:00] {
  "name": "Ping",
  "value": "SUCCESS (  548364 usec)\n",
  "check_id": ""
} 
```

### DIFF
--- a/internal/factsengine/gatherers/saphostctrl.go
+++ b/internal/factsengine/gatherers/saphostctrl.go
@@ -10,10 +10,15 @@ import (
 )
 
 const (
-	SapHostCtrlGathererName               = "saphostctrl"
-	saphostCtrlListInstancesParsingRegexp = `^\s+Inst Info\s*:\s*([^-]+?)\s*-\s*(\d+)\s*-\s*([^,]+?)` +
-		`\s*-\s*(\d+),\s*patch\s*(\d+),\s*changelist\s*(\d+)$`
-	saphostCtrlPingParsingRegexp = `(SUCCESS|FAILED) \( *(\d+) usec\)`
+	SapHostCtrlGathererName = "saphostctrl"
+)
+
+// nolint:gochecknoglobals
+var (
+	saphostCtrlListInstancesParsingRegexp = regexp.MustCompile(`^\s+Inst Info\s*` +
+		`:\s*([^-]+?)\s*-\s*(\d+)\s*-\s*([^,]+?)` +
+		`\s*-\s*(\d+),\s*patch\s*(\d+),\s*changelist\s*(\d+)$`)
+	saphostCtrlPingParsingRegexp = regexp.MustCompile(`(SUCCESS|FAILED) \( *(\d+) usec\)`)
 )
 
 // nolint:gochecknoglobals
@@ -30,7 +35,7 @@ var (
 	}
 	SapHostCtrlUnsupportedFunction = entities.FactGatheringError{
 		Type:    "saphostctrl-webmethod-error",
-		Message: "requested webmethod not whitelisted",
+		Message: "requested webmethod not supported",
 	}
 	SapHostCtrlParseError = entities.FactGatheringError{
 		Type:    "saphostctrl-parse-error",
@@ -74,17 +79,17 @@ func (g *SapHostCtrlGatherer) Gather(factsRequests []entities.FactRequest) ([]en
 
 func handleWebmethod(
 	executor utils.CommandExecutor,
-	incomingCommand string,
+	webMethod string,
 ) (entities.FactValue, *entities.FactGatheringError) {
-	webMethodHandler, ok := whitelistedWebmethods[incomingCommand]
+	webMethodHandler, ok := whitelistedWebmethods[webMethod]
 
 	if !ok {
-		gatheringError := SapHostCtrlUnsupportedFunction.Wrap(incomingCommand)
+		gatheringError := SapHostCtrlUnsupportedFunction.Wrap(webMethod)
 		log.Error(gatheringError)
 		return nil, gatheringError
 	}
 
-	saphostctlOutput, commandError := executeSapHostCtrlCommand(executor, incomingCommand)
+	saphostctlOutput, commandError := executeSapHostCtrlCommand(executor, webMethod)
 	if commandError != nil {
 		return nil, commandError
 	}
@@ -104,16 +109,15 @@ func executeSapHostCtrlCommand(executor utils.CommandExecutor, command string) (
 }
 
 func parsePing(commandOutput string) (entities.FactValue, *entities.FactGatheringError) {
-	re := regexp.MustCompile(saphostCtrlPingParsingRegexp)
 	pingData := map[string]entities.FactValue{}
 
-	matches := re.FindStringSubmatch(commandOutput)
+	matches := saphostCtrlPingParsingRegexp.FindStringSubmatch(commandOutput)
 	if len(matches) < 2 {
 		return nil, SapHostCtrlParseError.Wrap(commandOutput)
 	}
 
 	pingData["status"] = &entities.FactValueString{Value: matches[1]}
-	pingData["elapsed"] = &entities.FactValueString{Value: matches[2]}
+	pingData["elapsed"] = entities.ParseStringToFactValue(matches[2])
 
 	result := &entities.FactValueMap{Value: pingData}
 
@@ -121,14 +125,13 @@ func parsePing(commandOutput string) (entities.FactValue, *entities.FactGatherin
 }
 
 func parseInstances(commandOutput string) (entities.FactValue, *entities.FactGatheringError) {
-	re := regexp.MustCompile(saphostCtrlListInstancesParsingRegexp)
 	lines := strings.Split(commandOutput, "\n")
 	instances := []entities.FactValue{}
 
 	for _, line := range lines {
 		instance := map[string]entities.FactValue{}
-		if re.MatchString(line) {
-			fields := re.FindStringSubmatch(line)
+		if saphostCtrlListInstancesParsingRegexp.MatchString(line) {
+			fields := saphostCtrlListInstancesParsingRegexp.FindStringSubmatch(line)
 			if len(fields) < 6 {
 				return nil, SapHostCtrlParseError.Wrap(commandOutput)
 			}
@@ -136,9 +139,9 @@ func parseInstances(commandOutput string) (entities.FactValue, *entities.FactGat
 			instance["system"] = &entities.FactValueString{Value: fields[1]}
 			instance["instance"] = &entities.FactValueString{Value: fields[2]}
 			instance["hostname"] = &entities.FactValueString{Value: fields[3]}
-			instance["revision"] = &entities.FactValueString{Value: fields[4]}
-			instance["patch"] = &entities.FactValueString{Value: fields[5]}
-			instance["changelist"] = &entities.FactValueString{Value: fields[6]}
+			instance["sapkernel"] = entities.ParseStringToFactValue(fields[4])
+			instance["patch"] = entities.ParseStringToFactValue(fields[5])
+			instance["changelist"] = entities.ParseStringToFactValue(fields[6])
 
 			instances = append(instances, &entities.FactValueMap{Value: instance})
 		}
--- a/internal/factsengine/gatherers/saphostctrl.go
+++ b/internal/factsengine/gatherers/saphostctrl.go
@@ -2,42 +2,56 @@ package gatherers
 
 import (
 	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/agent/pkg/factsengine/entities"
+	"github.com/trento-project/agent/pkg/utils"
 )
 
 const (
 	SapHostCtrlGathererName = "saphostctrl"
 )
 
+// nolint:gochecknoglobals
+var (
+	SapHostCtrlCommandError = entities.FactGatheringError{
+		Type:    "saphostctrl-cmd-error",
+		Message: "error executing saphostctrl command",
+	}
+)
+
 type SapHostCtrlGatherer struct {
-	executor CommandExecutor
+	executor utils.CommandExecutor
 }
 
 func NewDefaultSapHostCtrlGatherer() *SapHostCtrlGatherer {
-	return NewSapHostCtrlGatherer(Executor{})
+	return NewSapHostCtrlGatherer(utils.Executor{})
 }
 
-func NewSapHostCtrlGatherer(executor CommandExecutor) *SapHostCtrlGatherer {
+func NewSapHostCtrlGatherer(executor utils.CommandExecutor) *SapHostCtrlGatherer {
 	return &SapHostCtrlGatherer{
 		executor: executor,
 	}
 }
 
-func (g *SapHostCtrlGatherer) Gather(factsRequests []FactRequest) ([]Fact, error) {
-	facts := []Fact{}
+func (g *SapHostCtrlGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
 	log.Infof("Starting saphostctrl facts gathering process")
 
 	for _, factReq := range factsRequests {
+		var fact entities.Fact
 		// TODO: The instance number "00" could be an additional argument in the future
 		version, err := g.executor.Exec(
 			"/usr/sap/hostctrl/exe/saphostctrl", "-nr", "00", "-function", factReq.Argument)
 		if err != nil {
-			// TODO: Decide together with Wanda how to deal with errors. `err` field in the fact result?
-			log.Errorf("Error running saphostctrl: %s", err)
+			gatheringError := SapHostCtrlCommandError.Wrap(factReq.Argument)
+			log.Error(gatheringError)
+			fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+		} else {
+			fact = entities.NewFactGatheredWithRequest(factReq, &entities.FactValueString{Value: (string(version))})
 		}
-		fact := NewFactWithRequest(factReq, string(version))
+
 		facts = append(facts, fact)
 	}
 
-	log.Infof("Requested saphostctrl facts gathered")
+	log.Infof("Requested %s facts gathered", SapHostCtrlGathererName)
 	return facts, nil
 }
--- a/internal/factsengine/gatherers/saphostctrl.go
+++ b/internal/factsengine/gatherers/saphostctrl.go
@@ -11,10 +11,12 @@ import (
 
 const (
 	SapHostCtrlGathererName               = "saphostctrl"
-	saphostCtrlListInstancesParsingRegexp = `^\s+Inst Info\s*:\s*([^-]+?)\s*-\s*(\d+)\s*-\s*([^,]+?)\s*-\s*(\d+),\s*patch\s*(\d+),\s*changelist\s*(\d+)$`
-	saphostCtrlPingParsingRegexp          = `(SUCCESS|FAILED) \( *(\d+) usec\)`
+	saphostCtrlListInstancesParsingRegexp = `^\s+Inst Info\s*:\s*([^-]+?)\s*-\s*(\d+)\s*-\s*([^,]+?)` +
+		`\s*-\s*(\d+),\s*patch\s*(\d+),\s*changelist\s*(\d+)$`
+	saphostCtrlPingParsingRegexp = `(SUCCESS|FAILED) \( *(\d+) usec\)`
 )
 
+// nolint:gochecknoglobals
 var whitelistedWebmethods = map[string]func(utils.CommandExecutor) (entities.FactValue, *entities.FactGatheringError){
 	"Ping":          handlePing,
 	"ListInstances": handleListInstances,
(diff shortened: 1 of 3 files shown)


nelsonkopliku (approved):
Looks good!

Then, sorry if I bother you with DSL examples :sweat_smile: yet I it helps me to keep a connection between the different aspects.

In this case a user would write this piece of yaml
```
facts:
    - name: my_saphostctrl_ping
      gatherer: saphostctrl
      argument: Ping
```
LGTM

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,39 @@
...
+
+const (
+	SapHostCtrlGathererName = "saphostctrl"
+)
+
+type SapHostCtrlGatherer struct {
+	executor CommandExecutor
+}
+
+func NewSapHostCtrlGatherer() *SapHostCtrlGatherer {
```
CDimonaco, line 15: In general, if we use DI we should inject all the dependency in the constructor, and bind them to the struct, in that case we should have an argument of type CommandExecutor and then bind it to the struct. This function now defeat the purpose of DI because it's a dependency but it's hard wired to Executor{}. the executor should be passed by outside as dependency.

In that way you can avoid to put also the test in the same package, and you could use the package_test package name, and inject the mock that you have already in place in the costructor function and create the Gatherer :D

CDimonaco (dismissed):
Just the DI thing, than LGTM

internal/factsengine/gatherers/saphostctrl_test.go
```diff
@@ -0,0 +1,109 @@
+package gatherers // nolint:dupl
```
dottorblaster, line 1: Shouldn't this be `gatherers_test`? Having troubles with package names usually means that we have to use some sort of dependency injection to inject the mock for some struct

dottorblaster (changes requested):
Just a comment, then we can merge

internal/factsengine/gatherers/saphostctrl_test.go
arbulu89: Nope. The file belongs to the same package

internal/factsengine/gatherers/saphostctrl.go
arbulu89: I understand your reasoning, but I don't really intend to have DI here to give users more options.

This implementation is only to enable the unit testing. I don't really want users messing up with this thinking they need to provide an executor, because they don't. I would even remove that dependency if I could test the function without mocking the `exec.Command` output.

For having a test in a different file. I think it is still totally possible, having their own mock. Using the constructor, being that simple, isn't that required in the unit test in my opinion. So you can simply create the struct as it is done in the test. Anyway, I don't want to open up the discussion whether test files should go in the same or other package in this PR.

CDimonaco: We can discuss about package testing in another place, we also have that thing tracked on the issue tracker, no worries :D 

The point is that DI is not just about give the "user" more choices, but also for us as developer being able to swap dependency when needed, even in tests, the only reason that test works in because it's in the same package and we exploit the fact the in go we can initialize in the same package a struct passing unexported fields, like the executor in this case, we need the constructor in order to initialize the struct when the struct is used in another package, the fact that we want to provide a "default dependency" like this case, is totally suitable, I get your point, maybe we can switch to a builder pattern approach, and we have the default executor (it's more elegant than an optional parameter, go does not have default or named parameters :( )

What do you thing?

arbulu89: >  test works in because it's in the same package
I don't get this point. Why do you say this? The tests don't even use the constructor, they declare the struct with the mocked executor. I guess we could every well do this in other "test" package, as the struct is exported. Am I missing something?
Or do you mean the fact that the `commandexecutor` is in the same package? Well, this is on purpose, if for some reason we needed to change this functionality to other place (I don't know why), we would need to pass the variable there, or simply import the new package and use it in the same way I guess.

I think we need to consider that this "dependency" was only included for testing, for anything else, so knowing this fact, we can play the cards we want.

> maybe we can switch to a builder pattern approach, and we have the default executor (it's more elegant than an optional parameter, go does not have default or named parameters :( )

Ok, let me try this option, removing the named parameter

arbulu89: @CDimonaco Would it be something like this?
https://github.com/trento-project/agent/pull/71/commits/f08bd3308badd1359a21534070f938adef5cdeca

(not entirely sure, maybe you meant something different hehe)

CDimonaco: The functional options pattern is very elegant and declarative, I think you will enjoy it! :D 

We can adopt this pattern in all constructors and expand them in the future, without breaking the general api, but only adding a new option while providing support for old api

CDimonaco: Nope I mean not using the constructor as now, and even without a parameter but build the constructor with a builder pattern or a functional options pattern, providing in the costructor a default implementation, but leaving the struct declaration as is, without embedding the Executor,  as last commit

arbulu89: We can do one thing. Let me merge as it was before adding that strange commit, and you could open a PR with an example on how to do it. All of us would learn more from that hehe

arbulu89: > This works, but it's not a real injection, because the only way that this works is because the test is in the same package and you can set an unexported property

Ok, now I got it. Thanks!

nelsonkopliku: Alright, the conversation went technical, that's good :smile: 

Just adding my 2 cents :smile: 
In my background I have also seen the tendency to inject deps via constructors, and more often than not those deps happen to be abstractions.

The alternative of not passing from a constructor injection is similar, for me at least, to setting a private property from outside its access context (even if we only do that in tests), breaking encapsulation, thus considered avoidable.

Now, this is my personal experience, which drives me towards a preference of injecting deps in the constructor.

Yet I am very eager to learn what is the go preferred way.

CDimonaco: @nelsonkopliku Yes, the construct injection is the way, in go you could do that even making the constructor with a functional options pattern approach, so you constructor has a default implementation, but you can pass as parameter to the costructor, variadic arguments that are function of a certain type that manipulates the struct you are building, ad apply some changes to it during the construction process.

This approach is really ideomatic and really declarative, but at the end of the day is always a constructor that receives dependency and options :D

arbulu89: @CDimonaco ,
@rtorrero Will work on this once this PR and other is merged. To apply this technique in all the related gatherers

rtorrero (approved):
For my side, all relevant things have already been addressed or discussed and agreed to be addressed in next PRs (including the DI).

LGTM

arbulu89:
@dottorblaster @dottorblaster Finally the requested changes rebased. Could you have a 2nd look?

rtorrero:
@arbulu89 @dottorblaster @nelsonkopliku I've updated the gatherer to match the latest change in the fact gathering process and improved a bit how the output looks. Take a look at the tests to see if you agree this is usable, and I'll go ahead and write the docs.

Basically, the changes in the output are that we now use a map with parsed output fields instead of the whole output as a single string.

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func (g *SapHostCtrlGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting saphostctrl facts gathering process")
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+
+		if !whitelisted()[factReq.Argument] {
```
arbulu89, line 55: We don't really need to run the `whitelisted` function all the times in the loop.
We could assign it to a variable before going in the loop

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,142 @@
...
+
+// nolint:gochecknoglobals
+var (
+	SapHostCtrlCommandError = entities.FactGatheringError{
+		Type:    "saphostctrl-cmd-error",
+		Message: "error executing saphostctrl command",
+	}
+	SapHostCtrlUnsupportedFunction = entities.FactGatheringError{
+		Type:    "saphostctrl-func-error",
+		Message: "requested function not whitelisted",
```
arbulu89, line 26: They are call `webmethods`. Maybe we should use this term instead of function.
What do you think?

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,142 @@
...
+	log.Infof("Starting saphostctrl facts gathering process")
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+
+		if !whitelisted()[factReq.Argument] {
+			gatheringError := SapHostCtrlUnsupportedFunction.Wrap(factReq.Argument)
+			log.Error(gatheringError)
+			fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+		} else {
```
arbulu89, line 59: This `else` could be avoided simply adding a `continue` statement to the previous `if` condition.
It makes the code easier to read.
You just need to have the `facts = append(facts, fact)` line twice, which I see as a good tradeoff

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,142 @@
...
+			saphostctlOutput, err := g.executor.Exec("/usr/sap/hostctrl/exe/saphostctrl", "-function", factReq.Argument)
+			if err != nil {
+				gatheringError := SapHostCtrlCommandError.Wrap(factReq.Argument)
+				log.Error(gatheringError)
+				fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+				facts = append(facts, fact)
+				continue
+			}
+			switch factReq.Argument {
```
arbulu89, line 68: I think we could improve the readability here a lot with something like this:
```
parsingFunctions := map[string]func(string) entities.Fact{
	"Ping": parsePing,
	"ListInstances": parseListInstances,
}
...
// to use 
parsingFunctions[factReq.Argument](saphostctlOutput)
```

internal/factsengine/gatherers/saphostctrl_test.go
```diff
@@ -0,0 +1,129 @@
...
```
arbulu89, line 1: Use `gatherers_test`

internal/factsengine/gatherers/saphostctrl_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (suite *SapHostCtrlTestSuite) SetupTest() {
+	suite.mockExecutor = new(utilsMocks.CommandExecutor)
+}
+
+func (suite *SapHostCtrlTestSuite) TestSapHostCtrlGather() {
+	suite.mockExecutor.On("Exec", "/usr/sap/hostctrl/exe/saphostctrl", "-function", "Ping").Return(
+		[]byte("SUCCESS (  543341 usec)\n"), nil)
+	suite.mockExecutor.On("Exec", "/usr/sap/hostctrl/exe/saphostctrl", "-function", "ListInstances").Return(
+		[]byte(" Inst Info : PRD - 00 - myhost-vmhana02 - 753, patch 410, changelist 1908545\n"), nil)
```
arbulu89, line 29: I think we should test what happens when we have 2 instances, to look if the code loops fine.

internal/factsengine/gatherers/saphostctrl_test.go
```diff
@@ -0,0 +1,129 @@
...
+
```
arbulu89, line 30: I would add other scenario when there is not any instance, to see that it returns an empty map

arbulu89 (commented):
Hey @rtorrero ,
Some changes requested.
Besides that, I would say some paths of the code are not tested (some parsing errors, command error, etc)

internal/factsengine/gatherers/saphostctrl_test.go
rtorrero: Thanks for this, I realized it was wrongly processing multiple instances, Should be good now

rtorrero:
@arbulu89 I've addressed some of your concerns and marked them as "resolved" for others, I've come with slightly different approaches but those should also cover your concerns. Please take a look at let me know :+1:

internal/factsengine/gatherers/saphostctrl_test.go
nelsonkopliku: And we're probably missing a test for a failed `ListInstances` as well

nelsonkopliku (changes requested):
Looks good Ruben, thanks!

I noted down a few minor things and then for me we're good.

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested %s facts gathered", SapHostCtrlGathererName)
+	return facts, nil
+}
+
+func isWhitelisted(command string) bool {
```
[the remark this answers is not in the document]
arbulu89: More than that, I don't think we need a function for this, as it is used only in one place.
You should better put this line in the place where it is used

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+	return whitelistedWebmethods[incomingCommand](executor)
+}
+
+func handlePing(executor utils.CommandExecutor) (entities.FactValue, *entities.FactGatheringError) {
+	saphostctlOutput, commandError := executeSapHostCtrlCommand(executor, "Ping")
+	if commandError != nil {
+		return nil, commandError
+	}
+
+	parsedOutput, parsingError := parsePing(saphostctlOutput)
```
arbulu89, line 99: All of this is not needed. You should better do `return parsePing(saphostctlOutput)`.
Besides that, I don't personally like to have so many function (handle and parse).
Why don't you put the parsing code here?
It is not used anywhere. I makes reading more difficult as you need to scroll up and down many times to find the code.
I would put everything about the `Ping` here.
The same goes for the other commands

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+	if parsingError != nil {
+		return nil, parsingError
+	}
+
+	return parsedOutput, nil
+}
+
+func handleListInstances(executor utils.CommandExecutor) (entities.FactValue, *entities.FactGatheringError) {
+	saphostctlOutput, commandError := executeSapHostCtrlCommand(executor, "ListInstances")
```
arbulu89, line 108: Why don't we put this 
```
saphostctlOutput, commandError := executeSapHostCtrlCommand(executor, "ListInstances")
if commandError != nil {
	return nil, commandError
}
```
in the main `Gather` function?
You need to duplicate it otherwise in many places

internal/factsengine/gatherers/saphostctrl_test.go
```diff
@@ -0,0 +1,198 @@
...
+	suite.Run(t, new(SapHostCtrlTestSuite))
+}
+
+func (suite *SapHostCtrlTestSuite) SetupTest() {
+	suite.mockExecutor = new(utilsMocks.CommandExecutor)
+}
+
+func (suite *SapHostCtrlTestSuite) TestSapHostCtrlGatherListInstances() {
+	suite.mockExecutor.On("Exec", "/usr/sap/hostctrl/exe/saphostctrl", "-function", "ListInstances").Return(
+		[]byte(" Inst Info : S41 - 41 - s41app - 785, patch 50, changelist 2091708\n"+
```
arbulu89, line 28: You could maybe replace this by:
```
const output = ` Inst Info : S41 - 41 - s41app - 785, patch 50, changelist 2091708
 Inst Info : S41 - 40 - s41app - 785, patch 50, changelist 2091708
 Inst Info : HS1 - 11 - s41db - 753, patch 819, changelist 2069355`
```
The `+` thing looks ugly

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+		if re.MatchString(line) {
+			fields := re.FindStringSubmatch(line)
+			if len(fields) < 6 {
+				return nil, SapHostCtrlParseError.Wrap(commandOutput)
+			}
+
+			instance["system"] = &entities.FactValueString{Value: fields[1]}
+			instance["instance"] = &entities.FactValueString{Value: fields[2]}
+			instance["hostname"] = &entities.FactValueString{Value: fields[3]}
+			instance["revision"] = &entities.FactValueString{Value: fields[4]}
```
arbulu89, line 165: shouldn't `revision`, `patch` and `changelist` be integers?

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+	re := regexp.MustCompile(saphostCtrlPingParsingRegexp)
+	pingData := map[string]entities.FactValue{}
+
+	matches := re.FindStringSubmatch(commandOutput)
+	if len(matches) < 2 {
+		return nil, SapHostCtrlParseError.Wrap(commandOutput)
+	}
+
+	pingData["status"] = &entities.FactValueString{Value: matches[1]}
+	pingData["elapsed"] = &entities.FactValueString{Value: matches[2]}
```
arbulu89, line 142: Shouldn't `elapsed` be an integer?

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+
+// nolint:gochecknoglobals
+var (
+	SapHostCtrlCommandError = entities.FactGatheringError{
+		Type:    "saphostctrl-cmd-error",
+		Message: "error executing saphostctrl command",
+	}
+	SapHostCtrlUnsupportedFunction = entities.FactGatheringError{
+		Type:    "saphostctrl-webmethod-error",
+		Message: "requested webmethod not whitelisted",
```
arbulu89, line 33: I would rephrase this to `requested webmethod is not supported`

internal/factsengine/gatherers/saphostctrl_test.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	suite.NoError(err)
+	suite.ElementsMatch(expectedResults, factResults)
+}
+
+func (suite *SapHostCtrlTestSuite) TestSapHostCtrlGatherError() {
+	suite.mockExecutor.On("Exec", "/usr/sap/hostctrl/exe/saphostctrl", "-function", "Ping").Return(
+		[]byte("FAILURE\n"), nil)
+	suite.mockExecutor.On("Exec", "/usr/sap/hostctrl/exe/saphostctrl", "-function", "Pong").Return(
```
arbulu89, line 153: What about an scenario where the command fails.
Here we have a
- parsing error
- unsupported command

So we miss the command error test

arbulu89 (commented):
Hey @rtorrero ,
I find the code difficult to follow with so many functions that are not meant to be functions (in my opinion), as they are not reused. And this creates the need to duplicate code many times.
I don't know I have the feeling you aimed to an extreme tailoring for functions. To makes reading difficult, as you need to scroll up and down constantly.

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		gatheringError := SapHostCtrlCommandError.Wrap(err.Error())
+		log.Error(gatheringError)
+		return "", gatheringError
+	}
+
+	return string(saphostctlOutput), nil
+}
+
+func parsePing(commandOutput string) (entities.FactValue, *entities.FactGatheringError) {
```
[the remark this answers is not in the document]
nelsonkopliku: not relevant anymore

internal/factsengine/gatherers/saphostctrl.go
```diff
@@ -0,0 +1,176 @@
...
+
+	pingData["status"] = &entities.FactValueString{Value: matches[1]}
+	pingData["elapsed"] = &entities.FactValueString{Value: matches[2]}
+
+	result := &entities.FactValueMap{Value: pingData}
+
+	return result, nil
+}
+
+func parseInstances(commandOutput string) (entities.FactValue, *entities.FactGatheringError) {
```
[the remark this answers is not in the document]
nelsonkopliku: not relevant anymore

internal/factsengine/gatherers/saphostctrl_test.go
rtorrero: I'm not sure I agree with this one, IMHO it looks worse because you can't indent the code properly :/

arbulu89: But that's a common thing in golang for this case.
Anyway, up to you. I'm fine with both options

arbulu89 (commented):
@rtorrero 
Super perfect from my side! I cannot approve since I opened the PR XD

nelsonkopliku (approved):

PR #73, opened 2022-08-08
Add cloudevents usage to wrap the inter-service messages

arbulu89:
Implement the `cloudevents` and `jsonschema` usage in the Agent side.

- It gets cloudevents, and depending on the type, it runs some operations. If the coming message content_type in the queue is `application/cloudevents+json`, it will unmarshall the code as a cloud event. After that, depending on the event_type in this cloud event, it will run the facts gathering execution. There is not any other supported operation by now. 
- The coming requests are validated against proper json schemas depending on the event type. This `jsonschema` code should be moved together with the schemas to the new `contracts` project at some point. The `json_schemas` go within the factsengine folder because embed requires so. You can't embed files that are not located within your folder. I guess the contracts project will fix this.
- It sends the gathered facts in other cloudevents message

The names of the events are just some initial options.

PD: This PR doesn't aim to change the rabbitmq adapter (which maybe it should be renamed to `amqp` as it does Wanda) adding a more "generic" way of using the broker. We could work on that in a parallel work of course


internal/factsengine/json_schema/trento.checks.v1.FactsGathered.schema.json
```diff
@@ -0,0 +1,45 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
nelsonkopliku, line 2: This is basically the same one that we have in [wanda](https://github.com/trento-project/wanda/blob/main/priv/json_schema/trento.checks.v1.FactsGathered.schema.json), right?
And these two will converge in the contracts repo.

internal/factsengine/gathering_test.go
```diff
@@ -0,0 +1,196 @@
...
+
+func NewErrorGatherer() *ErrorGatherer {
+	return &ErrorGatherer{}
+}
+
+func (s *ErrorGatherer) Gather(_ []gatherers.FactRequest) ([]gatherers.Fact, error) {
+	return []gatherers.Fact{}, fmt.Errorf("kabum!") //nolint
+}
+
+func (suite *GatheringTestSuite) TestGatheringGatherFacts() {
```
nelsonkopliku, line 64: The test name sound a bit redundant :sweat_smile: 

This is the happy path, right? What about `TestFactsGathering|TestSuccessfulFactsGathering` or similar?

internal/factsengine/adapters/rabbitmq.go
```diff
@@ -52,13 +53,14 @@ func (r *RabbitMQAdapter) Unsubscribe() error {
...
 }
 
-func (r *RabbitMQAdapter) Listen(agentID string, handle func([]byte) error) error {
+func (r *RabbitMQAdapter) Listen(agentID string, handle func(string, []byte) error) error {
 	return r.consumer.StartConsuming(
 		func(d rabbitmq.Delivery) rabbitmq.Action {
 			// TODO: Handle different kind of errors returning some sort of metadata
 			// so the applied action is potentially changed
-			err := handle(d.Body)
+			err := handle(d.ContentType, d.Body)
```
nelsonkopliku, line 61: If I understand it correctly, this means the content type `application/cloudevents+json` is expected in the amqp headers, and it will also be part of the body of the message.

An if I get it right it comes handy in order to avoid unmarshalling the body at this point.

It doesn't seem wanda puts that header, at the moment. We need to make the things click at a certain point :smile:

internal/factsengine/policy.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+	default:
+		return fmt.Errorf("Invalid event type: %s", eventType)
+	}
+	return nil
+}
+
+func handleContentType(contentType string, event []byte) (*cloudevents.Event, error) {
+	switch contentType {
+	case cloudevents.ApplicationCloudEventsJSON:
```
nelsonkopliku, line 41: Interesting, this would allow the agent to ingest different types of integration events next to the cloudevents.

Let's keep an eye on this and see how far to go.

arbulu89:
@CDimonaco @nelsonkopliku @rtorrero Closing this PR as we will remake it using the new contracts repository functionalities

PR #74, opened 2022-08-08
Implement `sbd dump` gatherer

rtorrero:
This PR implements the gatherer for the `sbd dump` command. 

Notice that I had to tweak a bit the `FindMatches` function from the utils, as it was replacing all the trailing white spaces that are returned from the command output with `_`

Also note that this gatherer requires two arguments:
 - The /dev/<id> device identifier for the SBD device in question
 - The key that the fact request is trying to collect.
 
The current FactRequest definition only allows for a single `argument` string so I encoded two arguments separated by a `:` character. e.g. the watchdog timeout for device /dev/sdj is requested by sending the following `FactRequest.Argument`:
`/dev/sdj:Timeout (watchdog)`


internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/agent/internal/utils"
+)
+
+const (
+	SBDDumpFactKey = "SBD_dump"
```
arbulu89, line 12: why `SBD_dump` hehe?
I would name it `sbd_dump` or `sbddump`

arbulu89, line 12: By the way, in other gatherers I have renamed this to `*GathererName` as the `FactKey` was something older, and it doesn't make much sense.
It would be `SBDDumpGathererName`

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (s *SBDDumpGatherer) Gather(factsRequests []FactRequest) ([]Fact, error) {
+	facts := []Fact{}
+	log.Infof("Starting %s facts gathering process", SBDDumpFactKey)
+
+	for _, factReq := range factsRequests {
+		args := strings.Split(factReq.Argument, ":")
+		SBDDump, err := s.executor.Exec(
+			"sbd", "dump", "-d", args[0], "dump", "-b")
```
arbulu89, line 32: I don't understand this command. The command should be `sbd -d /dev/sdc dump`. What is this `dump -b` at the end.

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,50 @@
...
+		executor: Executor{},
+	}
+}
+
+func (s *SBDDumpGatherer) Gather(factsRequests []FactRequest) ([]Fact, error) {
+	facts := []Fact{}
+	log.Infof("Starting %s facts gathering process", SBDDumpFactKey)
+
+	for _, factReq := range factsRequests {
+		args := strings.Split(factReq.Argument, ":")
```
arbulu89, line 30: We should do some error handling here to check the arguments length, or at least put some comment saying that this is temporary solution, and that the argument usage will change soonish

internal/factsengine/gatherers/sbddump_test.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func TestSBDDumpTestSuite(t *testing.T) {
+	suite.Run(t, new(SBDDumpTestSuite))
+}
+
+func (suite *SBDDumpTestSuite) TestSBDDumpGathererMissingFact() {
+	mockExecutor := new(mocks.CommandExecutor)
+
+	mockOutputFile, _ := os.Open("../../../test/fixtures/gatherers/sbddump.output")
```
arbulu89, line 24: If we are using the same dump in all the tests, we could have this in the test suite, and store it in a variable, so we don't need to open the file every check

internal/utils/utils.go
```diff
@@ -86,7 +86,7 @@ func FindMatches(pattern string, text []byte) map[string]interface{} {
 	r := regexp.MustCompile(pattern)
 	values := r.FindAllStringSubmatch(string(text), -1)
 	for _, match := range values {
-		key := strings.Replace(match[1], " ", "_", -1) //nolint
+		key := strings.Replace(strings.TrimSpace(match[1]), " ", "_", -1) //nolint
```
arbulu89, line 89: I don't really get this change. If we have spaces in our match, it means that the pattern we provided is wrong, and we should fix that, instead of changing the functionality here

arbulu89: I think this patter should make the job:
```
(?m)^(\S+(?: \S+)*)\s*:\s(\S*)$
```
(It allows words and 2 words with a unique space in between)
https://regex101.com/r/OCmlXk/1

arbulu89 (changes requested):
Thanks @rtorrero ,
Small changes requested. We can merge after they are merged.
I recommend you to test the gatherer in a real target. I think the current code wouldn't work hehe

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,50 @@
...
+
+const (
+	SBDDumpFactKey = "SBD_dump"
+)
+
+type SBDDumpGatherer struct {
+	executor CommandExecutor
+}
+
+func NewSBDDumpGatherer() *SBDDumpGatherer {
```
CDimonaco, line 19: The constructor function should have the dependency as parameter, in order to inject that. So a parameter of CommandExecutor type is needed. This way the constructor function wires a dependency that cannot be changed so it defeats the purpose of having a dependency injected and declared.

With this change you can inject the mock in tests, and do the test in the separate test package, and initialize the gatherer with the proper function, injecting the mocked dependency instead of initialize the struct without the constructor and use the private scope of the same package :D

CDimonaco (changes requested):
Only the DI thing in the constructor, and LGTM

internal/factsengine/gatherers/sbddump.go
rtorrero: I'd just go for this one as-is and switch to using DI with CommandExecutor next week on a follow-up PR, wdyt @CDimonaco ?

internal/factsengine/gatherers/sbddump_test.go
rtorrero: If it's ok, let's just keep it like this as I will make changes around this next week when I add DI for the CommandExecutor

rtorrero:
@CDimonaco I think it should be fine now; I'll care care in a followup PR of using DI for the commandexecutor

internal/factsengine/gatherers/sbddump_test.go
```diff
@@ -0,0 +1,139 @@
...
+	mockExecutor.On("Exec", "sbd", "dump", "-d", "/dev/sdj", "dump").Return(mockOutput, nil)
+
+	c := &SBDDumpGatherer{
+		executor: mockExecutor,
+	}
+
+	factRequests := []FactRequest{
+		{
+			Name:     "nonexistant_timeout",
+			Gatherer: "SBD_dump",
```
nelsonkopliku, line 35: shouldn't this be `sbd_dump` as the gatherer name? :thinking:

rtorrero:
@CDimonaco I'm guessing we can ignore those linter warnings for now and address them in a separate PR, right?

nelsonkopliku:
Disclaimer: Wall of text :sweat_smile: 

This touches checks related also with this PR #77  

[1.3.5](https://github.com/trento-project/runner/tree/main/runner/ansible/roles/checks/1.3.5) SBD watchdog timeout is set to `expectedValue` (60)
[1.3.6](https://github.com/trento-project/runner/tree/main/runner/ansible/roles/checks/1.3.6) SBD msgwait timeout value is two times the watchdog timeout

Let's consider this **1.3.5 SBD watchdog timeout is set to `expectedValue`**, [here's](https://github.com/trento-project/runner/blob/main/runner/ansible/roles/checks/1.3.5/tasks/main.yml) what we have now in ansible, which I understand it as: 
all the devices defined in `SBD_DEVICE` must have the same expected watchdog timeout.

Which from an implementation perspective requires being able to extract the devices in `SBD_DEVICE`, in sbd config file, and for each device run an `sbd -d ${device} dump` to get the needed values: `Timeout (watchdog)` in this case.
Check passes if `Timeout (watchdog)` value is the same on all the devices and that value is the expected one.

I find it useful to have in mind the resulting DSL of the checks we have, which also drives implementation, so usually I ask myself: how could this be represented with the DSL?

Let's start from what we are implementing:
- the `sbd_config` gatherer #77 - gathering `SBD_DEVICE`s
- `sbd_dump` gatherer in current PR - gathering whatever from `sbd dump`

How could these fit together?
```yaml
# SBD watchdog timeout is set to `expectedValue`
facts:
  - name: sbd_devices
    gatherer: sbd_config # this gatherer could return a list of devices? ["device1", "device2", "deviceN"]
    argument: SBD_DEVICE
  - name: sbd_watchdog_timeout 
    gatherer: sbd_dump
   # this gatherer would need input possibly from the previous fact definition
    argument: "/sbd/device:Timeout (watchdog)" # this would need a list of devices from previous fact
    # becoming actually something like this
    argument: $sbd_devices # reference to a fact, the $ is just an example, it can be anything
    # what would this gatherer return? a map with all the timeouts per device? 
expectations:
  - name: sbd_watchdog_timeout_correctly_set
    expect: sbd_watchdog_timeout == 60 # this expectation might be different given the fact's shape
```
This would require supporting facts/gatherers dependencies, where a gatherer's output is input for another gatherer.
I find this a very interesting feature that we could reason about in terms of value, complexity, priority, but this is not the topic.

What I am really thinking about is that, at least for the involved checks, It looks like some implementation details are leaking.
ie. _get these sbd devices which then would be used to get the whatever thing from the `sbd dump` command..._

In fact, as of now, both the involved checks use `sbd_dump` always in combination with a previous sbd devices extraction, as a mean to define the real fact (usually something being required to be the same on all devices).

In what scenarios would we use it as an _independent_ gatherer? 
```yaml
  - name: some_fact_from_sbd_dump 
    gatherer: sbd_dump
    argument: "${device}:${something}"
```
I am just wondering if I am missing other usages. 
Nevertheless, even if a direct usage of the `sbd_dump` as a gatherer would be meaningful, the leaking implementation detail issue remains, imho. And also the inter-facts dependency, which we could defer.

Again, I believe that having in mind the resulting DSL helps a lot.
To do that we need to understand the check, and define what's the facts and what's the expectations.

Check 1.3.5: **SBD watchdog timeout is set to `expectedValue`**

Option 1
**Fact:** SBD Watchdog Timeout

**Gatherer:** `sbd_watchdog_timeout`
This gatherer does not need any argument.
Returns either the detected watchdog timeout, if it is the same on **all** the devices, otherwise an error.
```
# pseudocode
theValue | Error {
    type: "sbd_mismatching_watchdog_timeout"
    messge: "whatever useful message",
    context: {
        "device1": "30",
        "device2": "50",
        "device3": "60",
    }
}
```

**Expectation:** SBD Watchdog Timeout is set to `expectedValue`

**Resulting yaml**
```yaml
# SBD watchdog timeout is set to `expectedValue`
facts:
  - name: sbd_watchdog_timeout
    gatherer: sbd_watchdog_timeout
expectations:
  - name: sbd_watchdog_timeout_correctly_set
    expect: sbd_watchdog_timeout == 60
```
---
Option 2
**Fact:** Collection of SBD Watchdog Timeout for all the devices

**Gatherer:** `sbd_watchdog_timeout`
This gatherer does not need any argument.
Returns either the collection of the detected watchdog timeouts, for all devices, or an error if any of the timeout extractions failed (aka `sbd dump` command fails)
```
# pseudocode
[
    {
        device: "device1",
        watchdogTimeout: "30"
    }
    {
        device: "device2",
        watchdogTimeout: "50"
    }
    {
        device: "device3",
        watchdogTimeout: "60"
    }
] | Error {
    type: "sbd_unable_to_extract_watchdog_timeout" <- could this case also be an sbd_mismatching_watchdog_timeout?
    messge: "whatever useful message",
    context: {
        # whatever useful context for the error
    }
}
```

**Expectation:** SBD Watchdog Timeout is the **same on all devices** AND that **value is set to `expectedValue`**

**Resulting yaml**
```yaml
# SBD watchdog timeout is set to `expectedValue`
facts:
  - name: sbd_watchdog_timeouts
    gatherer: sbd_watchdog_timeout
expectations:
  - name: sbd_watchdog_timeout_correctly_set
    expect: allEqual(sbd_watchdog_timeouts) && uniqueValue(sbd_watchdog_timeouts) == 60
#           ^^^^^^^^                           ^^^^^^^^^^^^                          ^^
#           if all the values in the list are equal, and that value is the expected one

# variants of the expectation
# allEqual(sbd_watchdog_timeouts) && sbd_watchdog_timeouts[0] == 60
# allEqualTo(sbd_watchdog_timeouts, 60)
# sorry for the naming
```

Any of these two options to me are 
- simpler and clearer from a DSL perspective
- encapsulate implementation details by composing different steps (extract devices + run the `sbd dump` command ...)
- simpler from an implementation perspective by allowing us to defer inter-facts dependencies

What do you think @trento-project/developers?

internal/factsengine/gatherers/sbddump_test.go
```diff
@@ -0,0 +1,139 @@
...
+			Argument: "/dev/sdj:Header version",
+		},
+		{
+			Name:     "sector_size",
+			Gatherer: "sbd_dump",
+			Argument: "/dev/sdj:Sector size",
+		},
+		{
+			Name:     "watchdog_timeout",
+			Gatherer: "sbd_dump",
```
nelsonkopliku, line 72: How would the checks involving `sbd_dump` (1.3.5 and 1.3.6) be represented with the DSL?
Let's say [1.3.5](https://github.com/trento-project/runner/blob/main/runner/ansible/roles/checks/1.3.5/tasks/main.yml) SBD watchdog timeout is set to `expectedValue` (60)

Translating the current implementation to 
> all the devices defined in `SBD_DEVICE` must have the same expected watchdog timeout.
>
it seems that we have to deal with lists of sbd devices and possibly their respective dumped values, in order to be able to express that.

How does that fit all together?

rtorrero: We still haven't decided this, but it could indeed require a bit more thought. This is how it would be specified now (with an alternative if we are not happy with having to provide the SBD device to the gatherer):
```
facts:
  - name: sbd_watchdog_timeout 
    gatherer: sbd_dump
    argument: "/sbd/device:Timeout (watchdog)"
    ## Alternatively, with a bit of work we could have:
    # argument: "Timeout (watchdog)"
    ## to gather the watchdog timeout for all devices (and automatically return an error on the gatherer when they don't match each other)

expectations:
  - name: sbd_watchdog_timeout_correctly_set
    expect: sbd_watchdog_timeout == 60 
  	## or if using the alternative, the same but in plural:
 	# - name: all_sbd_watchdog_timeouts_correctly_set
 	expect: all_sbd_watchdog_timeouts == 60 
```

rtorrero:
For the sake of cleaning up a bit, can we agree to close this PR?

angelabriel:
When can we expect the availability of the `sbd_dump` gatherer for our check rewrite activities? Any timeline available?

fabriziosestito:
> When can we expect the availability of the `sbd_dump` gatherer for our check rewrite activities? Any timeline available?

@angelabriel this is planned for the current sprint but it could take longer due to the xmas vacations

angelabriel:
> @angelabriel this is planned for the current sprint but it could take longer due to the xmas vacations

Great news, thanks. And no rush, I simply want to organize my own work.

PR #77, opened 2022-08-10
Sbd gatherer

nelsonkopliku:
This PR adds a first version of the SBD Config Gatherer, responsible to extract interesting values from sbd config file.

Here's the involved checks, so far
[1.3.1](https://github.com/trento-project/runner/tree/main/runner/ansible/roles/checks/1.3.1) SBD_PACEMAKER value is correctly set to `yes` in SBD configuration
[1.3.2](https://github.com/trento-project/runner/tree/main/runner/ansible/roles/checks/1.3.2) SBD_STARTMODE is set to `always`

And here's how the respective DSLs would look like
```
# 1.3.1 `SBD_PACEMAKER` value is correctly set in SBD configuration
facts:
  - name: sbd_pacemaker
    gatherer: sbd_config
    argument: SBD_PACEMAKER
expectations:
  - name: sbd_pacemaker_correctly_configured
    expect: sbd_pacemaker == yes

# 1.3.2 SBD_STARTMODE is set to always
facts:
  - name: sbd_startmode
    gatherer: sbd_config
    argument: SBD_STARTMODE
expectations:
  - name: sbd_startmode_correctly_configured
    expect: sbd_startmode == always
```

Open points:
- what should be returned to wanda in case we weren't able to parse the config file?
- what should be returned to wanda in case a value (fact) does not exist in sbd config file?

At the moment, since we yet don't have agreement I just used a string, as soon as we find an agreement I will refactor.


arbulu89 (commented):
@nelsonkopliku 

I wouldn't increase the golang version so lightly.
Specially in a PR that doesn't aim to that.

I think such a decision should be studied, and at least decided among the team, as it looks like something important.

My 2 cents hehe

internal/factsengine/gatherers/sbd.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/trento-project/agent/internal/cluster"
+)
+
+const (
+	SBDConfigGathererName = "sbd_config"
+	UndefinedSBDConfig    = "trento.checks.sbd.errors.undefined_configuration"
+)
+
+type SBDGatherer struct {
+	configFile string
```
arbulu89, line 15: Maybe we could use afero to make this testable. It is used in other gatherers like this one. As reference: https://github.com/trento-project/agent/pull/78

internal/factsengine/gatherers/sbd.go
```diff
@@ -0,0 +1,52 @@
...
+	return &SBDGatherer{
+		configFile,
+	}
+}
+
+func (g *SBDGatherer) Gather(factsRequests []FactRequest) ([]Fact, error) {
+	gatheredFacts := []Fact{}
+	log.Infof("Starting SBD Facts gathering")
+
+	conf, err := godotenv.Read(g.configFile)
```
arbulu89, line 32: We could use the `utils.FindMatches` function here, which does exactly the same, and we avoid other dependency.
It could be:
```
conf := utils.FindMatches(`(?m)^(\w+)=(\S[^#\s]*)`, fileContent)
```

arbulu89: In fact, as you already import `cluster`, it has a method named `getSBDConfig`, which does already the same

arbulu89 (commented):
And some comments about the sbd gatherer itself.
Take them as opinions, as the current implementation works perfectly fine as well

nelsonkopliku:
> I wouldn't increase the golang version so lightly. Specially in a PR that doesn't aim to that.
> 
> I think such a decision should be studied, and at least decided among the team, as it looks like something important.
> 

Thanks Xabi, I get your point and I will keep go at version 1.16 as it is out of scope for this PR. 
Here's the commit https://github.com/trento-project/agent/pull/77/commits/2243d4632973a5d387d427784e691d2663fc6fdd

While fixing the CI linting issues coming from main, I realized that as of go 1.16 (or later) the `"io/ioutil"` package is deprecated and its functionalities split around between `io` and `os` packages, and I usually try to keep languages/tools to the most up to date version, in order to not get stuck in unsupported/obsolete tech cages :sweat_smile: 

So, I asked for help yesterday evening to fix those linting issues and, among the other things bumping the go version seemed to be acceptable and viable.

Now, since I didn't yet have the chance to share with you as well:
what would block us from bumping go to 1.19 it in the next PR :smile:? 
What should we take into account that we aren't?
The only thing that comes to my mind is that obs/ibs do not have go 1.19, is that the case?

Let's also involve the rest of the team:
@CDimonaco @fabriziosestito @dottorblaster @rtorrero what do you think about bumping go to version 1.19? In another PR of course :smile:

dottorblaster:
@nelsonkopliku you have my sword

internal/factsengine/gatherers/sbd.go
nelsonkopliku: The absence of afero is just because of the usage of the https://github.com/joho/godotenv library, which under the hood uses simply `os.Whatever(...)`.
Since we are providing the configuration file path as a dependency, testing works pretty well, imho:
- at runtime the configuration file is the proper `cluster.SBDConfigPath` aka `cluster.SBDConfigPath`
- during tests the proper fake file is provided `../../../test/sbd_config`

The point is: if I can instruct the code to use a specific file (actual one or test one), I don't mind using the straight `os` file related functions, because I will just provide the correct file path in the correct scenario.
This thing, in this case at least, allows us to change the file we use, not the complete filesystem abstraction we are using, and I find it ok. Does this make sense?

internal/factsengine/gatherers/sbd.go
nelsonkopliku: This is interesting. I absolutely didn't see/remember that function :sweat_smile: 

Anyway, could you help me understand the issue with adding the https://github.com/joho/godotenv dependency?

A middle ground that I would find useful would be the other way around:
refactoring what we currently do in `cluster.getSBDConfig` by using `godotenv`, export that funcion and use it in the gatherer.

Less code, less things to test and maintain :stuck_out_tongue: 
Anyway the library looks quite stable and suitable for the kind of file we are processing.

What do you think?

rtorrero:
> > I wouldn't increase the golang version so lightly. Specially in a PR that doesn't aim to that.
> > I think such a decision should be studied, and at least decided among the team, as it looks like something important.
> 
> Thanks Xabi, I get your point and I will keep go at version 1.16 as it is out of scope for this PR. Here's the commit [2243d46](https://github.com/trento-project/agent/commit/2243d4632973a5d387d427784e691d2663fc6fdd)
> 
> While fixing the CI linting issues coming from main, I realized that as of go 1.16 (or later) the `"io/ioutil"` package is deprecated and its functionalities split around between `io` and `os` packages, and I usually try to keep languages/tools to the most up to date version, in order to not get stuck in unsupported/obsolete tech cages sweat_smile
> 
> So, I asked for help yesterday evening to fix those linting issues and, among the other things bumping the go version seemed to be acceptable and viable.
> 
> Now, since I didn't yet have the chance to share with you as well: what would block us from bumping go to 1.19 it in the next PR smile? What should we take into account that we aren't? The only thing that comes to my mind is that obs/ibs do not have go 1.19, is that the case?
> 
> Let's also involve the rest of the team: @CDimonaco @fabriziosestito @dottorblaster @rtorrero what do you think about bumping go to version 1.19? In another PR of course smile

I se go 1.18 in `devel:languages:go/go` in OBS. I could easily test this next week; my fork is also linked to my own OBS packages where I can test easily if everything builds fine there. +1 for updating to 1.19 but also +1 for doing it maybe in a separate PR

dottorblaster (approved):
I think this is mergeable :+1:

rtorrero (approved):
Lgtm!

PR #78, opened 2022-08-10
Implement /etc/hosts file gatherer

rtorrero:
This PR:
 - Changes a bit the corosyncconf gatherer to use dependency injection for the interaction with the filesystem. This improves the testability
 - Implements a gatherer that collects facts about the /etc/hosts file. Given a hostname as argument, it will return a list of IPs that the hostname resolves to.
 - Takes care about some minor complains by the linter

### DIFF
--- a/internal/factsengine/gatherers/hostsfile.go
+++ b/internal/factsengine/gatherers/hostsfile.go
@@ -2,7 +2,6 @@ package gatherers
 
 import (
 	"bufio"
-	"errors"
 	"fmt"
 	"os"
 	"regexp"
@@ -21,7 +20,7 @@ const (
 )
 
 var (
-	HostsEntryCompiled = regexp.MustCompile(parsingRegexp)
+	hostsEntryCompiled = regexp.MustCompile(parsingRegexp)
 )
 
 // nolint:gochecknoglobals
@@ -70,19 +69,15 @@ func (s *HostsFileGatherer) Gather(factsRequests []entities.FactRequest) ([]enti
 
 	for _, factReq := range factsRequests {
 		var fact entities.Fact
-		var found bool
 
-		for hostname, ip := range hostsFileMap.Value {
-			if hostname == factReq.Argument {
-				fact = entities.NewFactGatheredWithRequest(factReq, ip)
-				facts = append(facts, fact)
-				found = true
-				break
-			}
+		if ip, found := hostsFileMap.Value[factReq.Argument]; found {
+			fact = entities.NewFactGatheredWithRequest(factReq, ip)
+		} else {
+			gatheringError := HostsEntryNotFoundError.Wrap(factReq.Argument)
+			log.Error(gatheringError)
+			fact = entities.NewFactGatheredWithError(factReq, gatheringError)
 		}
-		if !found {
-			return nil, errors.New(HostsEntryNotFoundError.Error())
-		}
+		facts = append(facts, fact)
 	}
 
 	log.Infof("Requested /etc/hosts file facts gathered")
@@ -117,12 +112,12 @@ func hostsFileToMap(lines []string) (*entities.FactValueMap, error) {
 	var paramsMap = make(map[string]string)
 
 	for _, line := range lines {
-		match := HostsEntryCompiled.FindStringSubmatch(line)
+		match := hostsEntryCompiled.FindStringSubmatch(line)
 
 		if match == nil {
 			return nil, fmt.Errorf("invalid hosts file structure")
 		}
-		for i, name := range HostsEntryCompiled.SubexpNames() {
+		for i, name := range hostsEntryCompiled.SubexpNames() {
 			if i > 0 && i <= len(match) {
 				paramsMap[name] = match[i]
 			}
@@ -130,8 +125,8 @@ func hostsFileToMap(lines []string) (*entities.FactValueMap, error) {
 		hostnames := strings.Fields(paramsMap["hostnames"])
 
 		for _, hostname := range hostnames {
-			if i, found := hostsFileMap[hostname]; found {
-				if ipsByHostname, ok := i.(*entities.FactValueList); ok {
+			if ip, found := hostsFileMap[hostname]; found {
+				if ipsByHostname, ok := ip.(*entities.FactValueList); ok {
 					ipsByHostname.Value = append(ipsByHostname.Value, &entities.FactValueString{Value: paramsMap["ip"]})
 				} else {
 					return nil, fmt.Errorf("casting error while mapping ips to hosts")
--- a/internal/factsengine/gatherers/hostsfile.go
+++ b/internal/factsengine/gatherers/hostsfile.go
@@ -42,7 +42,7 @@ var (
 )
 
 type HostsFileGatherer struct {
-	hostsFile string
+	hostsFilePath string
 }
 
 func NewDefaultHostsFileGatherer() *HostsFileGatherer {
@@ -50,14 +50,14 @@ func NewDefaultHostsFileGatherer() *HostsFileGatherer {
 }
 
 func NewHostsFileGatherer(hostsFile string) *HostsFileGatherer {
-	return &HostsFileGatherer{hostsFile: hostsFile}
+	return &HostsFileGatherer{hostsFilePath: hostsFile}
 }
 
 func (s *HostsFileGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
 	facts := []entities.Fact{}
 	log.Infof("Starting /etc/hosts file facts gathering process")
 
-	hostsFile, err := readHostsFileByLines(s.hostsFile)
+	hostsFile, err := readHostsFileByLines(s.hostsFilePath)
 	if err != nil {
 		return nil, HostsFileError.Wrap(err.Error())
 	}
@@ -90,17 +90,23 @@ func readHostsFileByLines(filePath string) ([]string, error) {
 		return nil, err
 	}
 
-	defer hostsFile.Close()
+	defer func() {
+		err := hostsFile.Close()
+		if err != nil {
+			log.Error(err)
+		}
+	}()
 
 	fileScanner := bufio.NewScanner(hostsFile)
 	fileScanner.Split(bufio.ScanLines)
 	var fileLines []string
 
 	for fileScanner.Scan() {
-		if strings.HasPrefix(fileScanner.Text(), "#") || fileScanner.Text() == "" {
+		scannedLine := fileScanner.Text()
+		if strings.HasPrefix(scannedLine, "#") || scannedLine == "" {
 			continue
 		}
-		fileLines = append(fileLines, fileScanner.Text())
+		fileLines = append(fileLines, scannedLine)
 	}
 
 	return fileLines, nil
--- a/internal/factsengine/gatherers/hostsfile.go
+++ b/internal/factsengine/gatherers/hostsfile.go
@@ -17,12 +17,11 @@ const (
 	HostsFilePath       = "/etc/hosts"
 	ipMatchGroup        = "ip"
 	hostnamesMatchGroup = "hostnames"
+	parsingRegexp       = `(?m)(?P<` + ipMatchGroup + `>\S+)\s+(?P<` + hostnamesMatchGroup + `>.+)`
 )
 
 var (
-	parsingRegexp      = `(?m)(?P<` + ipMatchGroup + `>\S+)\s+(?P<` + hostnamesMatchGroup + `>.+)`
 	HostsEntryCompiled = regexp.MustCompile(parsingRegexp)
-	//HostsEntryCompiled = regexp.MustCompile(`(?m)(?P<ip>\S+)\s+(?P<hostnames>.+)`)
 )
 
 // nolint:gochecknoglobals
@@ -73,7 +72,7 @@ func (s *HostsFileGatherer) Gather(factsRequests []entities.FactRequest) ([]enti
 		var fact entities.Fact
 		var found bool
 
-		for hostname, ip := range hostsFileMap {
+		for hostname, ip := range hostsFileMap.Value {
 			if hostname == factReq.Argument {
 				fact = entities.NewFactGatheredWithRequest(factReq, ip)
 				facts = append(facts, fact)
@@ -112,8 +111,9 @@ func readHostsFileByLines(filePath string) ([]string, error) {
 	return fileLines, nil
 }
 
-func hostsFileToMap(lines []string) (map[string][]string, error) {
-	var hostsFileMap = make(map[string][]string)
+func hostsFileToMap(lines []string) (*entities.FactValueMap, error) {
+	var hostsFileMap = make(map[string]entities.FactValue)
+
 	var paramsMap = make(map[string]string)
 
 	for _, line := range lines {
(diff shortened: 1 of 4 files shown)


internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/trento-project/agent/internal/utils"
+)
+
+const (
+	HostsFactKey  = "hosts"
+	HostsFilePath = "/etc/hosts"
```
CDimonaco, line 17: If we don't use them outside the package we can leave them unexported, both the constants

CDimonaco (changes requested):
Just the const things, and then LGTM, all good!

internal/factsengine/factsengine_test.go
```diff
@@ -409,7 +408,7 @@ func (suite *FactsEngineTestSuite) TestFactsEngineLoadPlugins() {
 
 	engine := &FactsEngine{ // nolint
 		factGatherers: map[string]gatherers.FactGatherer{
-			gatherers.CorosyncFactKey: gatherers.NewCorosyncConfGatherer(),
+			gatherers.CorosyncCmapCtlFactKey: gatherers.NewCorosyncCmapctlGatherer(),
```
arbulu89, line 411: Nothing wrong with it, but why this change?

internal/factsengine/factsengine.go
```diff
@@ -21,17 +22,19 @@ type FactsEngine struct {
 }
 
 func NewFactsEngine(agentID, factsEngineService string) *FactsEngine {
+	fileSystem := afero.NewOsFs()
```
arbulu89, line 25: So at the end we go with the DI hehe
If this is the case @rtorrero , could you take care of changing the rest of the gatherers that use the `CommandExecutor` so they use the same technique, and we pass the as variables.
I'm not a big fan of that, but hey, it looks like everyone else prefers this way.
You could do it of course in other PR

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/trento-project/agent/internal/utils"
+)
+
+const (
+	HostsFactKey  = "hosts"
```
arbulu89, line 16: I would call this `HostsGathererName` . The `FactKey` thing go obsolete, I would say.

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,74 @@
...
+	for _, factReq := range factsRequests {
+		factValue := searchMapByValue(hostsMap, factReq.Argument)
+		if len(factValue) < 1 {
+			continue
+		}
+		fact := NewFactWithRequest(factReq, searchMapByValue(hostsMap, factReq.Argument))
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested Hosts file facts gathered")
```
arbulu89, line 55: Maybe change to `Requested /etc/hosts file facts gathered` to align with the first logging message

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	hostsMap := utils.FindMatches(`(?m)^([^#|\s]+)\s([^#\n]+)#?.*$`, hostsFileContents)
+
+	for _, factReq := range factsRequests {
+		factValue := searchMapByValue(hostsMap, factReq.Argument)
+		if len(factValue) < 1 {
+			continue
+		}
+		fact := NewFactWithRequest(factReq, searchMapByValue(hostsMap, factReq.Argument))
```
arbulu89, line 51: We could use here the `factValue`, right?
No need to call the function back again.

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return facts, err
+	}
+
+	hostsMap := utils.FindMatches(`(?m)^([^#|\s]+)\s([^#\n]+)#?.*$`, hostsFileContents)
+
+	for _, factReq := range factsRequests {
+		factValue := searchMapByValue(hostsMap, factReq.Argument)
+		if len(factValue) < 1 {
+			continue
```
arbulu89, line 49: Is this the piece of code where we should return an failed fact? Could we put a `TODO` comment?

internal/factsengine/gatherers/hostsfile_test.go
```diff
@@ -0,0 +1,114 @@
...
+	confFile, _ := ioutil.ReadAll(testFile)
+	err := afero.WriteFile(suite.fileSystem, "/etc/hosts", confFile, 0644)
+	assert.NoError(suite.T(), err)
+	c := NewHostsFileGatherer(suite.fileSystem)
+
+	factRequests := []FactRequest{
+		{
+			Name:     "hosts_localhost",
+			Gatherer: "hosts",
+			Argument: "localhost",
```
arbulu89, line 42: I would put the `CheckID` in the `FactRequest` entry, so we test that the same `CheckID` is present in the result

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+		fact := NewFactWithRequest(factReq, searchMapByValue(hostsMap, factReq.Argument))
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested Hosts file facts gathered")
+	return facts, nil
+}
+
+func searchMapByValue(m map[string]interface{}, value string) []string {
```
arbulu89, line 59: The function looks a bit cryptic, isn't it? hehe

arbulu89 (changes requested):
Nice @rtorrero ,
Some minor changes asked.

About the DI, as commented in one comment below, as looks like everybody prefers the option to add it in the constructor, (even other PRs started having such a things), could you work in other PR to use the `CommandExecutor` this way.
Thanks

internal/factsengine/factsengine_test.go
rtorrero: Basically because the CorosyncCmapctlGatherer doesn't require to pass any parameter to the constructor. But I guess that is changing soon enough :-)

internal/factsengine/gatherers/hostsfile.go
rtorrero: Totally, I moved this last minute and forgot.

internal/factsengine/gatherers/hostsfile.go
rtorrero: I hope it looks a bit better now :-)

internal/factsengine/gatherers/hostsfile.go
rtorrero: I wouldn't add a TODO there as I think it could be an option to just return no results for a fact request. But we need to discuss that and come back to this.

rtorrero:
Most comments have been addressed, should be good to merge now but let me know :-)

internal/factsengine/gatherers/hostsfile_test.go
```diff
@@ -0,0 +1,120 @@
...
+			CheckID:  "check1",
+		},
+		{
+			Name:     "hosts_somehost",
+			Gatherer: "hosts",
```
nelsonkopliku, line 47: Once again, how does this fit in the DSL representation of the involved check (1.5.1)?

rtorrero: facts:
  - name: nodes_hosts_file_resolution
    gatherer: hosts_file
    argument: "node01:node02"

expectations:
  - name: nodes_resolvable_using_hosts_file
	# not sure how to express that the resulting list should have at least 2 results, just a suggestion
    expect: len(nodes_hosts_file_resolution) > 1

rtorrero: With the new expression evaluation in place, we can now use e.g. `array.some(|x| x == expected_value)` to check for the presence of an IP in the obtained facts, so it should be fine. If we don't know such IP we can just check for the length of the resulting array.

rtorrero:
I've re-used this old PR and rewritten a big chunk of it. I think it should be fine now!

@nelsonkopliku @CDimonaco @arbulu89 would you mind checking again?

rtorrero:
Im meanwhile adjusting to include the changes from  https://github.com/trento-project/agent/pull/133

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,138 @@
...
+	hostsFileMap, err := hostsFileToMap(hostsFile)
+	if err != nil {
+		return nil, HostsFileDecodingError.Wrap(err.Error())
+	}
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+		var found bool
+
+		for hostname, ip := range hostsFileMap {
```
arbulu89, line 76: Do you need to loop here?
Being a map, you could access with the key
```
if ip, found : = hostsFileMap(factReq.Argument); found {
...
```

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,138 @@
...
+			if hostname == factReq.Argument {
+				fact = entities.NewFactGatheredWithRequest(factReq, ip)
```
arbulu89, line 78: You will need to wrap the `ip` in the new format:
```
&FactValueString{Value: ip}
```

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,138 @@
...
+				facts = append(facts, fact)
+				found = true
+				break
+			}
+		}
+		if !found {
+			return nil, errors.New(HostsEntryNotFoundError.Error())
```
arbulu89, line 85: You must continue looping here even though you have 1 failure. There might be other facts for other checks.
You could have something like:
```
gatheringError := HostsEntryNotFoundError.Wrap(factReq.Argument)
log.Error(gatheringError)
fact = entities.NewFactGatheredWithError(factReq, gatheringError)
```

internal/factsengine/gatherers/hostsfile_test.go
```diff
@@ -0,0 +1,101 @@
+package gatherers
```
arbulu89, line 1: Could we have other test where the provided argument is not found? to test the abnormal path of the lookup

arbulu89 (changes requested):
Hey @rtorrero ,
Did you consider using some package like the nexts to load and parse the hosts file:
https://github.com/jaytaylor/go-hostsfile
https://github.com/txn2/txeh/blob/master/txeh.go

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -128,11 +128,22 @@ func hostsFileToMap(lines []string) (map[string][]string, error) {
 			}
 		}
 		hostnames := strings.Fields(paramsMap["hostnames"])
+
 		for _, hostname := range hostnames {
-			hostsFileMap[hostname] = append(hostsFileMap[hostname], paramsMap["ip"])
+			if i, found := hostsFileMap[hostname]; found {
```
CDimonaco, line 133: just a minor thing, I really appreciate the fact that you are using proper variable name, it's all more readable, maybe we could use ip, instead of i, seems more aligned with the other names, what do you think?

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -17,12 +17,11 @@ const (
 	HostsFilePath       = "/etc/hosts"
 	ipMatchGroup        = "ip"
 	hostnamesMatchGroup = "hostnames"
+	parsingRegexp       = `(?m)(?P<` + ipMatchGroup + `>\S+)\s+(?P<` + hostnamesMatchGroup + `>.+)`
 )
 
 var (
-	parsingRegexp      = `(?m)(?P<` + ipMatchGroup + `>\S+)\s+(?P<` + hostnamesMatchGroup + `>.+)`
 	HostsEntryCompiled = regexp.MustCompile(parsingRegexp)
```
CDimonaco, line 24: we use that regexp in other packages? Otherwise we could make that unexported

internal/factsengine/gatherers/hostsfile_test.go
rtorrero: It's already indirectly tested in the `TestHostsFileIgnoresCommentedHosts` test, but I can add a specific test as I noticed it may be not 100% clear.

rtorrero: In the end... I've decided not to add it, as it's basically a 1:1 copy of that tests. Let me know if it's ok

rtorrero:
> Hey @rtorrero , Did you consider using some package like the nexts to load and parse the hosts file: https://github.com/jaytaylor/go-hostsfile https://github.com/txn2/txeh/blob/master/txeh.go

I did check https://github.com/jaytaylor/go-hostsfile  but I discarded it as it just allows for reverse lookups (get the hostname from the IP).

Just looked at `txeh`, it should work but it's a bit overkill as it looks like it's aimed towards handling /etc/hosts file rather than fetching data fom it.

internal/factsengine/gatherers/hostsfile.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	HostsEntryNotFoundError = entities.FactGatheringError{
+		Type:    "hosts-file-value-not-found",
+		Message: "requested field value not found in /etc/hosts file",
+	}
+)
+
+type HostsFileGatherer struct {
+	hostsFile string
```
CDimonaco, line 45: I think that in general we could and we should use afero when possibile, in that case I think we could drop it, we could inject the path and use something like a fixture which we are doing right know in tests.

I just suggest to rename the hostsFile parameter here and in constructor function to hostsFilePath, is less misleading and we know right that it's the path and not the file content.

CDimonaco:
I left just one comment, but LGTM, good job!

internal/factsengine/gatherers/hostsfile.go
fabriziosestito: +1 to rename this prior merging

arbulu89 (approved):
Hey @rtorrero ,
I still think that we should use something like the `txnh` library. Or at least take reference on the code.
But if the tests are happy i'm happy as well.

PD:
The `txeh` code reference:
https://github.com/txn2/txeh/blob/master/txeh.go#L255
https://github.com/txn2/txeh/blob/master/txeh.go#L293

It works as simple as:
```
hs, err := txeh.NewHostsDefault()
found, address, _ := hs.HostAddressLookup("linux-afj9")
```

internal/factsengine/gatherers/hostsfile.go
rtorrero: Aight :+1:

rtorrero:
> Hey @rtorrero , I still think that we should use something like the `txnh` library. Or at least take reference on the code. But if the tests are happy i'm happy as well.
> 
> PD: The `txeh` code reference: https://github.com/txn2/txeh/blob/master/txeh.go#L255 https://github.com/txn2/txeh/blob/master/txeh.go#L293
> 
> It works as simple as:
> 
> ```
> hs, err := txeh.NewHostsDefault()
> found, address, _ := hs.HostAddressLookup("linux-afj9")
> ```

Let's defer adding a dependency for now, since we already have the code. If at any point in the future we find the need to e.g. auto remediate a "missing entry in the /etc/hosts file" check, we can always add `txeh` as a dependency and simplify this gatherer.

PR #84, opened 2022-08-29
Work on FIXMEs part 1

arbulu89:
First round of fixes and refactoring of the project code `FIXME` entries.

More to come, but I wanted to keep them small.

Next:
- Continue with the remaining FIXMEs
- Use proper testing suites in all tests
- Move to individual test packages

PR #85, opened 2022-08-30
Move the individual unit test function to suites

arbulu89:
Move the existing individual unit test to test suites.

PD: The sapsystem.go tests are dealt in: https://github.com/trento-project/agent/pull/84


rtorrero (approved):
Green light from me, GJ @arbulu89

PR #87, opened 2022-09-01
Fixmes round 2

arbulu89:
Second (and last by now) part of the FIXMEs battle.

It includes all the needed changes to fix all the current FIXMEs (which doesn't mean some might be unknown).

** I know, the PR might look super big, and it includes many file changes. But 95% of them are just copy/paste things, so it shouldn't need a thorough and deep review. In essence, the functionality reamins 100% the same than before.**
** I thought better to do it in a unique PR, otherwise, I would need several PRs, that would most probably consume more time on reviews**

Basically:
- Move the commandexecutor.go code to utils, as it is needed in more parts than only the fact gathering. This way, I reused this interface in all the needed places to later on mock it in the tests.
- It applies dependency injection in all the required code to avoid the usage of globals. Mainly, to run system commands.
- Define the variables declarative way instead of imperative way. This means that the fields are set when the variable is created, and not afterwards using assignments.
- Split the `sapsystem.go` code in smaller files.
- Remove many of the command execution mocks spreaded around the code, as we have a unique one now.
- Remove not used `mapstructure` annotations. They were old legacy annotations. We don't use this library anymore.
- Move the test fixture files to the correct fixtures folders instead of letting them hanging in the main `test` folder.


internal/sapsystem/sapsystem.go
```diff
@@ -298,24 +275,19 @@ func getDBAddress(system *SAPSystem) (string, error) {
...
+		return getUniqueIDHana(fs, sid)
 	case Application:
-		id, err = getUniqueIDApplication(system.SID)
+		return getUniqueIDApplication(executor, sid)
 	case DiagnosticsAgent:
-		id, err = getUniqueIDDiagnostics(fs)
+		return getUniqueIDDiagnostics(fs)
 	case Unknown:
-		id = "-"
+		return "-", nil
```
rtorrero, line 287: We could use a `default` here and avoid having the `return "-", nil` repeated again on the next line

rtorrero (approved):
Some minor comments left, overall LGTM

CDimonaco (approved):
LGTM

arbulu89:
Hey @rtorrero @CDimonaco ,
I'm closing this PR and open subsequent smaller PRs. The conflicts got difficult to fix, so let's try to go step by step

PR #88, opened 2022-09-02
Publish gathered facts using contract

arbulu89:
First interaction with the [contracts](https://github.com/trento-project/contracts) repository to publish the gathered facts.

Basically, it uses the contracts project to build the answer.

I have introduced the `policy.go` file to make it as similar as the wanda side (followed the old PR: https://github.com/trento-project/agent/pull/73). The facts gathering request logic will go in this file as well.

Some next step:
- Create a `entities` package within the `factsengine` to store the used structs (now in facts.go). We need to align the structs there with the ones defined in the contracts, following similar naming convention, fields, etc
- Unify naming conventions and rabbitmq usage values, like exchange names, queues, etc
- Do the same operation with the facts gathering request message, which is consumed by the agent. Pending to define the format of this message yet.


CDimonaco (approved):
LGTM

PR #93, opened 2022-09-06
Fact gathering error reply in results

arbulu89:
Implementation of the fact gathering errors reply in the results. 
When any error happens during a fact gathering, this is send in the `Error` field of the `FactGathered` struct.

This is the an example of the log and result:
```
INFO[2022-09-06 09:18:50] Starting facts gathering process             
INFO[2022-09-06 09:18:50] Starting Package versions facts gathering process 
ERRO[2022-09-06 09:18:51] fact gathering error: type: package_not_found - package not found: package other is not installed 
INFO[2022-09-06 09:18:51] Requested Package versions facts gathered    
INFO[2022-09-06 09:18:51] Requested facts gathered                     
INFO[2022-09-06 09:18:51] Publishing gathered facts to the checks engine service 
INFO[2022-09-06 09:18:51] Gathered facts published properly  
```

```
{
  "agent_id": "-",
  "execution_id": "-",
  "facts_gathered": [
    {
      "check_id": "check2",
      "name": "pacemaker_version",
      "value": "2.0.5+20201202.ba59be712"
    },
    {
      "check_id": "check3",
      "name": "corosync_version",
      "value": "2.4.5"
    },
    {
      "check_id": "check3",
      "error": {
        "message": "package not found: package other is not installed",
        "type": "package_not_found"
      },
      "name": "other_version",
      "value": null
    }
  ]
}
```

### DIFF
--- a/internal/factsengine/entities/facts_gathered.go
+++ b/internal/factsengine/entities/facts_gathered.go
@@ -3,13 +3,14 @@ package entities
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 
 	"github.com/pkg/errors"
 
 	contracts "github.com/trento-project/contracts/go/pkg/gen/entities"
 )
 
-type Error struct {
+type FactGatheringError struct {
 	Message string
 	Type    string
 }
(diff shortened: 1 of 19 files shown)


arbulu89:
> I was wondering about the gather method, with that approach what are the cases in which the function will return an error? I mean if we intecept and error in the command, and we return an entity populated with error and no error.
> 
> Do you think we can consider an approach where we populate the error in the function that calls the gather function, with the error returned from the gather function?

Actually, there is not any scenario for that. I was even thinking to remove that 2nd return value, but I didn't want to at least include it in this PR.

The thing about returning the fact discovery errors in that 2nd parameter, it is not that simple, because we might have different errors, we could even have a scenario where all the requests are returned with an error.

The fact that now multiple facts are gathered in the same execution makes it difficult. This was done in this way to avoid the need to get the `source of truth` multiple times (the less we run commands or read files, the better). In order to make what you say, we would need to split the task of getting the current data and finding the fact in this data.

arbulu89:
Closing this at the moment, as now as protobuf is used the implementation must be different

PR #95, opened 2022-09-07
Fact gathering requested

arbulu89:
Implemente the protobuf messaging system with the contracts coming from: https://github.com/fabriziosestito/contracts

**In order to make the change a bit smaller, I have moved the major part of the gatherers to a a _todo folder, so they are not compiled. Once we have the code working, adapting them should be easy**

In summary:
- Implement the usage of the protobuf contracts messaging
- Rename the domain struct so they match better with the messaging structure (even though there are internal structs to be consumed in the code)
- Implement the even handling policy code in the `policy.go` code
- Update the rabbitmq communication to use the agreed topics and queues
- The gathering code is exactly as before. I simply have moved the file to the new `gathering.go` code
- The `go.mod` file is a bit hacky to rename the project to the definitive landscape rather than the fabrizio's one. The golanci exception will go away at this point as well.


CDimonaco (approved):
LGTM

PR #97, opened 2022-09-16
Map the numeric strings as numbers to send the event

arbulu89:
Map the numeric strings as numbers in the event conversion. This way, if we gather some fact like `30000`, this will be automatically converted to a number.


CDimonaco (approved):
LGTM

fabriziosestito:
late to the party but great one!

PR #102, opened 2022-09-21
Extract command executor to utils package

arbulu89:
Extract `CommendExecutor` code to utils so it is accessible from other packages.

Imported from: https://github.com/trento-project/agent/pull/87


rtorrero (approved):
LGTM

PR #103, opened 2022-09-21
Use proper di in the sles subscription discovery

arbulu89:
Use proper dependency injection with the command executor in the sles subscriptions discovery code


rtorrero (approved):
LGTM!

PR #108, opened 2022-09-21
Remove mapstructure annotations

arbulu89:
`mapstructure` is not used since some time ago, so the annotations are not needed.


rtorrero (approved):
LGTM

PR #110, opened 2022-09-21
Detect Nutanix as underlying platform provider

nelsonkopliku:
This PR adds to the agent the capability to determine whether the OS is running on a Nutanix host.

As usual some specific content in the output of `demidecode` command is used determine these kind of information.
```bash
hana-g8-vm:~ # dmidecode | grep -i -E "nutanix|ahv"
	Version: nutanix-ahv-2.20220304.0.2429.el7
	Manufacturer: Nutanix
	Product Name: AHV 
```

In a nutshell
the `Instance` discovered during a `CloudDiscovery` can now have `nutanix` as value for the `Provider` property, and no `Metadata` would be available in this case.

Meh on the fact that we have this bare-metal thingy entangled with Cloud concepts, yet I find it wise to go like this, and track the debt. 
I am not sure we could do much better without scope creeping, and missing the point: supporting nutanix bare metal :metal:

### DIFF
--- a/internal/cloud/metadata.go
+++ b/internal/cloud/metadata.go
@@ -9,9 +9,10 @@ import (
 )
 
 const (
-	Azure = "azure"
-	AWS   = "aws"
-	GCP   = "gcp"
+	Azure   = "azure"
+	AWS     = "aws"
+	GCP     = "gcp"
+	Nutanix = "nutanix"
 	// DMI chassis asset tag for Azure machines, needed to identify wether or not we are running on Azure
 	// This is actually ASCII-encoded, the decoding into a string results in "MSFT AZURE VM"
 	azureDmiTag = "7783-7084-3265-9085-8269-3286-77"
@@ -89,6 +90,19 @@ func (i *Identifier) identifyGCP() (bool, error) {
 	return regexp.MatchString(".*Google.*", provider)
 }
 
+func (i *Identifier) identifyNutanix() (bool, error) {
+	log.Debug("Checking if the VM is running on Nutanix...")
+	output, err := i.executor.Exec("dmidecode")
+	if err != nil {
+		return false, err
+	}
+
+	dmidecodeContent := strings.TrimSpace(string(output))
+	log.Debugf("dmidecode output: %s", dmidecodeContent)
+
+	return regexp.MatchString("(?i)nutanix|ahv", dmidecodeContent)
+}
+
 func (i *Identifier) IdentifyCloudProvider() (string, error) {
 	log.Info("Identifying if the VM is running in a cloud environment...")
 
@@ -111,6 +125,13 @@ func (i *Identifier) IdentifyCloudProvider() (string, error) {
 	} else if result {
 		log.Infof("VM is running on %s", GCP)
 		return GCP, nil
+	}
+
+	if result, err := i.identifyNutanix(); err != nil {
+		return "", err
+	} else if result {
+		log.Infof("VM is running on %s", Nutanix)
+		return Nutanix, nil
 	}
 
 	log.Info("VM is not running in any recognized cloud provider")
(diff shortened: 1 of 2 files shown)


fabriziosestito (approved):
LGTM

internal/cloud/metadata.go
```diff
@@ -89,6 +90,19 @@ func (i *Identifier) identifyGCP() (bool, error) {
...
+	log.Debug("Checking if the VM is running on Nutanix...")
+	output, err := i.executor.Exec("dmidecode")
+	if err != nil {
+		return false, err
+	}
+
+	dmidecodeContent := strings.TrimSpace(string(output))
+	log.Debugf("dmidecode output: %s", dmidecodeContent)
+
+	return regexp.MatchString("(?i)nutanix|ahv", dmidecodeContent)
```
arbulu89, line 103: Maybe we should ask to Bernd what this `ahv` means. 
I know that he gave the result, but `dmidecode` has the option to only send the manufacturer with the `-s system-manufacturer` option. This way, we would only need to check if `Nutanix` is the output, like we do for AWS.
As we don't have the change to test ourselves, I guess that we will need to go with this.

arbulu89 (approved):
LGTM.
If we could have the option to have a nutanix test machine, we could improve the dmidecode command selection and the regular expression, but it is what it is hehe

internal/cloud/metadata_test.go
```diff
@@ -37,6 +37,16 @@ func dmidecodeGCP() []byte {
 	return []byte("Google")
 }
 
+func dmidecodeNutanix() []byte {
+	return []byte(`
+		SomeUselessProp: some-value-1.1.0
+
```
dottorblaster, line 43: Delete this empty line maybe?

dottorblaster (approved):
Tiny nitpick then LGTM

internal/cloud/metadata_test.go
nelsonkopliku: It was intended to make it clear that dmidecode output will very likely have some extra content not related to nutanix.

Yet, i don't mind removing it.

internal/cloud/metadata.go
```diff
@@ -89,6 +90,19 @@ func (i *Identifier) identifyGCP() (bool, error) {
 	return regexp.MatchString(".*Google.*", provider)
 }
 
+func (i *Identifier) identifyNutanix() (bool, error) {
+	log.Debug("Checking if the VM is running on Nutanix...")
```
rtorrero, line 94: I wouldn't assume this is a VM. An output such as "Checking if running under Nutanix" seems more appropriate.

internal/cloud/metadata.go
```diff
@@ -113,6 +127,13 @@ func (i *Identifier) IdentifyCloudProvider() (string, error) {
...
 	}
 
+	if result, err := i.identifyNutanix(); err != nil {
+		return "", err
+	} else if result {
+		log.Infof("VM is running on %s", Nutanix)
+		return Nutanix, nil
+	}
+
 	log.Info("VM is not running in any recognized cloud provider")
```
rtorrero, line 137: Same here

internal/cloud/metadata.go
rtorrero: `ahv` is the name of Nutanix own hypervisor (but from what i've read, they support multiple hypervisors). Not sure if in all cases it returns `Nutanix` as manufacturer though :/

rtorrero (changes requested):
Thanks @nelsonkopliku! Some small changes suggested in the output. I know there are already log messages exposing the same incorrect terms, we could fix them here on in a separate PR

rtorrero (approved):
Actually, just hit green once you change it, no need for re-request review :+1:

internal/cloud/metadata.go
nelsonkopliku: Referring not to a `system` 752cf74eb7e2d23bc9e5b6c04e18503c3a3bffc9

internal/cloud/metadata.go
nelsonkopliku: As above
Referring now to a `system` 752cf74eb7e2d23bc9e5b6c04e18503c3a3bffc9

PR #111, opened 2022-09-22
Move test fixture files

arbulu89:
Move fixture test files to the fixtures folder, to more meaningful places.
Besides that, I have changes the way of accessing the test files. The old `testing.go` (used with the `_ "github.com/trento-project/agent/test"`) import was creating some confusion as it changes the execution directory for all of the tests in the same package.
The new way is more explicit.


test/helpers/fixtures.go
```diff
@@ -0,0 +1,20 @@
+package helpers
+
+import (
+	"path"
+	"runtime"
+)
+
+var fixturesFolder = "" // nolint
```
dottorblaster, line 8: Why this nolint?

dottorblaster (commented):
Just a tiny comment

test/helpers/fixtures.go
arbulu89: Because `golangci-lint` doesn't like global variables, but this is a corner case. I could put `// nolint:gochecknoglobals` if that helps.

arbulu89: Changed and rebased

nelsonkopliku (approved):
Looks good.

CDimonaco (commented):
Just a note, the name of the function is GetFixtureFile, the name could be misleading because I expect that the function will actually return the file, could we call it GetFixturePath, or embed di os open in that and return the bytes/string

arbulu89:
> Just a note, the name of the function is GetFixtureFile, the name could be misleading because I expect that the function will actually return the file, could we call it GetFixturePath, or embed di os open in that and return the bytes/string?
> 
> What do you think?

Yes, you are right. Better name it `GetFixturePath`.
Thanks!

CDimonaco (approved):
LGTM

PR #112, opened 2022-09-22
Fix integration test to cancel properly listen function

arbulu89:
Fix the facts engine integration test. I noticed that even though it might be passing (even though it should be failing), it was happening because the checks were not done due to an incorrect cancel and wait of the main listen function


nelsonkopliku (approved):
Hope this helps my PR failing :smile: https://github.com/trento-project/agent/actions/runs/3104080121/jobs/5028174858

dottorblaster (approved):
LGTM

PR #113, opened 2022-09-22
Add kvm discovery

rtorrero:
This PR allows the agent to detect KVM as underlying platform, in a similar way as PR #110 does for `Nutanix`.

Instead of using `dmidecode`, we use here `systemd-detect-virt`, which reports `kvm` when executed on a virtual machine running on a KVM hypervisor:
Captura desde 2022-09-22 16-29-05

`dmidecode` was not giving enough details, only demostrating that QEMU was in use in the host machine:
Captura desde 2022-09-22 14-27-05

Notice that I've placed the `identifyKVM()` call after the `identifyNutanix()` call, as I suspect that `nutanix` is could also report `kvm` as it's hypervisor (AHV) is based on KVM. No way to confirm until we can test it.

This is a bit tricky though and is starting to show proof that we might want to restructure the code and separate the concepts of "Cloud Provider" and "virtualization provider" (the latter meaning "hypervisor"), as I could easily see `kvm` being detected in future providers we might include support for.

### DIFF
--- a/internal/cloud/metadata_test.go
+++ b/internal/cloud/metadata_test.go
@@ -46,6 +46,14 @@ func dmidecodeNutanix() []byte {
 	`)
 }
 
+func systemdDetectVirtKVM() []byte {
+	return []byte("kvm")
+}
+
+func systemdDetectVirtEmpty() []byte {
+	return []byte("none")
+}
+
 func dmidecodeEmpty() []byte {
 	return []byte("")
 }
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
Looks good.

internal/cloud/metadata_test.go
```diff
@@ -46,6 +46,14 @@ func dmidecodeNutanix() []byte {
 	`)
 }
 
+func systemdDetectVirtKVM() []byte {
+	return []byte("kvm")
+}
+
+func systemdDetectVirtEmpty() []byte {
+	return []byte("None")
```
arbulu89, line 54: The default output is `none` (small case) XD

arbulu89 (approved):
Good! I hope this new command is available in all suse systems hehe

fabriziosestito:
> Good! I hope this new command is available in all suse systems hehe

I think it is standard systemd

internal/cloud/metadata_test.go
rtorrero: true, thanks!

PR #115, opened 2022-09-27
Update Contracts with ContenType fetching from facade

CDimonaco:


.golangci.yaml
```diff
@@ -1,7 +1,7 @@
 linters-settings:
   gomoddirectives:
     replace-allow-list:
-      - github.com/trento-project/contracts
+      - github.com/trento-project/contracts/go
```
arbulu89, line 4: I think we can remove the `linters-settings` block as we don't have any replace anymore

internal/factsengine/mapper.go
```diff
@@ -130,7 +130,11 @@ func FactsGatheredToEvent(gatheredFacts entities.FactsGathered) ([]byte, error)
 		FactsGathered: facts,
 	}
 
-	eventBytes, err := events.ToEvent(&event, entities.FactsGathererdEventSource, uuid.New().String())
+	eventBytes, err := events.ToEvent(
+		&event,
+		events.WithSource(entities.FactsGathererdEventSource),
+		events.WithID(uuid.New().String()),
```
arbulu89, line 136: If we don't use the `WithID`, the uuid will be generated in the contract side in the same way.

arbulu89 (approved):
Great!
I would remove the `linters-settings` block in the golang ci.
Green light besides that

internal/factsengine/mapper.go
CDimonaco: Yes it's true, I wanted just to point it out as a matter of clarity, anyway we can discuss further on the sensible details of the contracts :D

.golangci.yaml
dottorblaster: I think @arbulu89 is right ._.

CDimonaco: done!

PR #118, opened 2022-10-11
Fix GHA obs jobs

stefanotorresi:
This PR changes the GHA setup so that it works with the latest changes in trento-project/continuous-delivery#13.

It also removes the need to use `obs-service-go_modules` with the RPM packaging (because upstream breaks it too frequently), and groups together all the static analysis jobs.

Same PR (more or less) will follow up on all the other repos with OBS CI jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,16 +195,17 @@ jobs:
     container:
       image: ghcr.io/stefanotorresi/continuous-delivery:fix-home-reference
       env:
-        GITHUB_OAUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         OSC_CHECKOUT_DIR: /tmp/trento-agent-package
-        HOME: /home/osc
       options: -u 0:0
     steps:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: configure OSC
-        run: /scripts/init_osc_creds.sh
+        run: |
+          mkdir -p $HOME/.config/osc
+          cp /home/osc/.config/osc/oscrc $HOME/.config/osc
+          /scripts/init_osc_creds.sh
       - name: prepare _service file
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -190,62 +189,75 @@ jobs:
...
-        # in order to add the current rolling release notes
-        if: github.event_name == 'release'
-        run: |
-          git config --global --add safe.directory /__w/agent/agent
-          osc checkout $OBS_PROJECT trento-agent trento-agent.changes
-          mv trento-agent.changes $FOLDER
-          VERSION=$(./hack/get_version_from_git.sh)
-          TAG=$(echo $VERSION | cut -f1 -d+)
-          hack/gh_release_to_obs_changeset.py $REPOSITORY -a shap-staff@suse.de -t $TAG -f $FOLDER/trento-agent.changes
+          cp /home/osc/.config/osc/oscrc $HOME/.config/osc
```
arbulu89, line 207: What is this `cp` for. Isn't `$HOME` `home/osc` already?

packaging/suse/_service
```diff
@@ -1,5 +1,5 @@
 <services>
-    <service name="tar_scm" mode="disabled">
+    <service name="tar_scm" mode="manual">
```
arbulu89, line 2: What is this for? Couldn't we just use `osc service un disabledrun`? What's the difference?
Was it deprecated?

arbulu89 (approved):
If obs maintainers are happy, I'm happy

packaging/suse/_service
stefanotorresi: yes, this was introduced recently I think. The practical difference is not much, but basically what was previously called `disabled` should always been `manual` because that's the intention: to run these services manually. This is now done via `osc service manualrun|mr`.

PR #120, opened 2022-10-18
Use google protobuf value in Fact message

fabriziosestito:


go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/spf13/viper v1.12.0
 	github.com/stretchr/testify v1.8.0
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
-	github.com/trento-project/contracts/go v0.0.0-20220923150043-8947bc9e21f5
+	github.com/trento-project/contracts/go v0.0.0-20221018071137-016f7ec1aed3
```
arbulu89, line 21: where did you get that commit? it looks something not going in the `main` branch.
This is the last: https://github.com/trento-project/contracts/commit/5a23ea08ca25c5cb2ea91691736ad814f52abc3e

arbulu89 (approved):
It works as an 1:1 translation. 
We would still need to send `struct` and `list` types, as we might start relaying in more complex test expression.
It can come in a future PR.
We should track it though

go.mod
fabriziosestito: yes i need to update since i merged the PR

PR #121, opened 2022-10-19
Refactor factengine integration test

fabriziosestito:
Moved factsengine inregration test inside factsengine package.

Use short flag to skip integration tests.

### DIFF
--- a/internal/factsengine/factsengine_integration_test.go
+++ b/internal/factsengine/factsengine_integration_test.go
@@ -1,6 +1,4 @@
-//go:build integration_test
-
-package integration_test
+package factsengine
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


internal/factsengine/factsengine_integration_test.go
```diff
@@ -1,6 +1,4 @@
-//go:build integration_test
-
-package integration_test
+package factsengine
```
arbulu89, line 1: We have started in many places the `_test` suffice. We should continue this way here:
`package factsengine_test`

arbulu89 (approved):
Thanks you @fabriziosestito 
Yes, it looks simpler.
I suggest we continue the `package_test` usage now on

CDimonaco:
LGTM, I agree with @arbulu89, we should implement the test as test package

internal/factsengine/factsengine_integration_test.go
fabriziosestito: @arbulu89 @CDimonaco should i move every other test in factsengine to the test package?

CDimonaco: If you want yes, It will help to keep consistency, If there are tests that cannot be moved (globals, testing of unexported thing etc.., just skip it, we will take them into another pr)

arbulu89: I wouldn't do it in this PR @fabriziosestito 
Let's merge this as it is already reviewed

PR #124, opened 2022-10-24
Use go-envparse in sbd gatherer

nelsonkopliku:
Follows up https://github.com/trento-project/agent/pull/122 by using [go-envparse](https://github.com/hashicorp/go-envparse) to parse sbd configuration.

Note: I couldn't really run tests locally :sweat_smile: some issues with the temporary `_todo` packaging.
I temporarily moved the files among the _real_ gatherers and tested that way (passed), then moved back in `_todo`

Low prio, we'll get back to gatherers at a certain point :smile:

### DIFF
--- a/internal/factsengine/gatherers/_todo/sbd.go
+++ b/internal/factsengine/gatherers/_todo/sbd.go
@@ -1,7 +1,11 @@
 package gatherers
 
 import (
-	"github.com/joho/godotenv"
+	"os"
+
+	"github.com/hashicorp/go-envparse"
+	"github.com/pkg/errors"
+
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
It looks good.
I guess that if we don't use the `godotenv` package anymore it should be removed from the go.mod and go.sum, right?

nelsonkopliku:
> It looks good. I guess that if we don't use the `godotenv` package anymore it should be removed from the go.mod and go.sum, right?

It's not there anymore, I guess it has been removed in some previous PR

PR #125, opened 2022-10-24
Add GroupID field to the FactsGathered event mapping

arbulu89:
Based on: https://github.com/trento-project/contracts/pull/26

I took advantage and cleaned some constants on the test using suite variables. Shared constants were confusing


CDimonaco (approved):
lgtm

PR #127, opened 2022-10-31
Match workflow name correctly for dependabot

arbulu89:
Fix `dependabot_auto_merge` to match the correct workflow name


arbulu89:
Duplicated: https://github.com/trento-project/agent/pull/126

PR #133, opened 2022-11-10
Add FactValue type

fabriziosestito:
Adds `FactValue` variant type and the possibility to return lists or maps from the gatherer.
Using variant types with sealed interfaces is not the most idiomatic thing in golang, but the `structpb` package from google uses the same trick to define the google protobuf value, and it saves us some runtime errors if the users returns types that we do not handle as `interface{}`.
The conversion to a protobuf message is straight forward by just using the `AsInterface()` method.

Existings gatehrers and plugins are refactored to comply with the new return type.

Also, I've added float and bool type and an utility function to parse a string to a number or a bool.

### DIFF
--- a/internal/factsengine/entities/fact_value.go
+++ b/internal/factsengine/entities/fact_value.go
@@ -1,0 +1,99 @@
+package entities
+
+import "strconv"
+
+// `FactValue` represents a dynamically typed value which can be either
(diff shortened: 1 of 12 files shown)


arbulu89 (approved):
Brilliant!
Love the change!

CDimonaco (approved):
LGTM, great work!

PR #137, opened 2022-11-16
Gather all hosts entries when no arg is provided

rtorrero:
This PR does a small change in the hosts file gatherer so that when _no_ hostname argument is provided, instead of returning a list of IP corresponding to that hostname, it returns a map with all of the hosts as key.

See the check for the `hosts_all` fact for more details.


internal/factsengine/gatherers/hostsfile.go
```diff
@@ -69,7 +69,11 @@ func (s *HostsFileGatherer) Gather(factsRequests []entities.FactRequest) ([]enti
 
 	for _, factReq := range factsRequests {
 		var fact entities.Fact
-
+		if factReq.Argument == "" {
```
nelsonkopliku, line 72: This needs to be double checked, it would probably be `nil`

internal/factsengine/gatherers/hostsfile_test.go
```diff
@@ -65,6 +70,36 @@ func (suite *HostsFileTestSuite) TestHostsFileBasic() {
 			}},
 			CheckID: "check3",
 		},
+		{
+			Name: "hosts_all",
+			Value: &entities.FactValueMap{
```
nelsonkopliku, line 75: How does this get transformed in our protobuf contracts?

From an elixir point of view I imagine wanda would get something like this

```elixir
%{
  "value" => %{
    "node1" => %{
      "value" => [
        "123.456.789",
        "1.1.1.1"
      ]
    }
  }
}
```
which in the context of an Evaluation becomes
```elixir
context = %{
  "facts" => %{
    "available_host_names" => %{
      "value" => %{
        "node1" => %{
          "value" => [
            "123.456.789",
            "1.1.1.1"
          ]
        }
      }
    }
  }
}
```

requiring to access to the fact via `.value`

```rust
["node1"].all(|cluster_node| cluster_node in facts.available_host_names.value)
```

nelsonkopliku (commented):
Alright, I left a couple of comments.

And here's a reference to related Wanda PR https://github.com/trento-project/wanda/pull/84/

Let's defer the two for a bit so we can address the open point with also the rest of the team.

internal/factsengine/gatherers/hostsfile.go
rtorrero: I checked it manually by not providing the `Argument` field in the fact request, and it seems that in such case, `factReq.Argument == ""` (in fact, the tests are passing). But I'll look again and figure out if it can be `nil`  with other inputs)

internal/factsengine/gatherers/hostsfile_test.go
rtorrero: I think so, but I invoke @fabriziosestito :smile:

internal/factsengine/gatherers/hostsfile.go
nelsonkopliku: Shame on me! Forgot that go indeed has "" as null value for strings.. my bad :sweat_smile:

arbulu89 (approved):

PR #139, opened 2022-11-16
Port systemd gatherer

arbulu89:
Port systemd gatherer to use the new FactValue object. The used argument is the name of the daemon to get the state.
It returns an `active/inactive` string. If the deamon is disabled or does not even exist, `inactive` is returned.
We could actually return more things as map if needed. All of [this](https://github.com/coreos/go-systemd/blob/d5623bf85e8e73ae6352f78ee6b55a287619dd4e/dbus/methods.go#L379)
 [Ref](https://github.com/coreos/go-systemd).

```
facts:
  - name: sbd_state
    gatherer: systemd
    argument: sbd
```

PD:
I have renamed some constants from other gatherers to use more meaningful names.

### DIFF
--- a/internal/factsengine/gatherers/corosyncconf.go
+++ b/internal/factsengine/gatherers/corosyncconf.go
@@ -14,8 +14,8 @@ import (
 )
 
 const (
-	CorosyncFactKey  = "corosync.conf"
-	CorosyncConfPath = "/etc/corosync/corosync.conf"
+	CorosyncConfGathererName = "corosync.conf"
+	CorosyncConfPath         = "/etc/corosync/corosync.conf"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


rtorrero (approved):
LGTM and it actually reminded me that we had agreed on changing the `FactKey` naming for something more understandable and that I need to change the testing package name with the `_test` in my last PRs too :sweat_smile:

arbulu89:
> LGTM and it actually reminded me that we had agreed on changing the `FactKey` naming for something more understandable and that I need to change the testing package name with the `_test` in my last PRs too sweat_smile

Aaa yes!
We should start looking on that in new pieces of code.
I forgot in the review...

nelsonkopliku (approved):

PR #140, opened 2022-11-16
Packageversion gatherer

rtorrero:
This PR adapts the packageversion gatherer to the latest changes to the fact gathering engine.


internal/factsengine/gatherers/packageversion.go
```diff
@@ -3,38 +3,50 @@ package gatherers
...
+	"github.com/trento-project/agent/internal/utils"
 )
 
 const (
 	PackageVersionGathererName = "package_version"
 )
 
+// nolint:gochecknoglobals
+var (
+	PackageVersionCMDError = entities.FactGatheringError{
```
arbulu89, line 15: Thie `CMD` looks horrible hehe
I would write `Command`

internal/factsengine/gatherers/packageversion.go
```diff
@@ -3,38 +3,50 @@ package gatherers
...
 	log.Infof("Starting Package versions facts gathering process")
 
 	for _, factReq := range factsRequests {
+		var fact entities.Fact
 		version, err := g.executor.Exec(
 			"rpm", "-q", "--qf", "%{VERSION}", factReq.Argument)
 		if err != nil {
-			// TODO: Decide together with Wanda how to deal with errors. `err` field in the fact result?
-			log.Errorf("Error getting version of package: %s", factReq.Argument)
+			gatheringError := PackageVersionCMDError.Wrap(factReq.Argument)
```
arbulu89, line 44: Could we log the error here?

internal/factsengine/gatherers/packageversion.go
```diff
@@ -3,38 +3,50 @@ package gatherers
...
+			fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+		} else {
+			fact = entities.NewFactGatheredWithRequest(factReq, entities.ParseStringToFactValue(string(version)))
```
arbulu89, line 47: I don't think we need the `ParseString..` here. We already know that this is a string.
I would put `&entities.FactValueString{Value: string(version)}` here

arbulu89 (changes requested):
Hey @rtorrero ,
It looks good! Some small changes requested.
PD: I forsee conflicts, so maybe rebase main in

arbulu89 (approved):

PR #141, opened 2022-11-17
Refactor sbd gatherer

nelsonkopliku:
This PR, resuming https://github.com/trento-project/agent/pull/77,  adds a first version of the SBD Config Gatherer, responsible to extract interesting values from sbd config file.

Here's the involved checks, so far
[1.3.1](https://github.com/trento-project/runner/tree/main/runner/ansible/roles/checks/1.3.1) SBD_PACEMAKER value is correctly set to `yes` in SBD configuration
[1.3.2](https://github.com/trento-project/runner/tree/main/runner/ansible/roles/checks/1.3.2) SBD_STARTMODE is set to `always`

And here's how the respective DSLs would look like
```
# 1.3.1 `SBD_PACEMAKER` value is correctly set in SBD configuration
facts:
  - name: sbd_pacemaker
    gatherer: sbd_config
    argument: SBD_PACEMAKER
expectations:
  - name: sbd_pacemaker_correctly_configured
    expect: sbd_pacemaker == "yes"

# 1.3.2 SBD_STARTMODE is set to always
facts:
  - name: sbd_startmode
    gatherer: sbd_config
    argument: SBD_STARTMODE
expectations:
  - name: sbd_startmode_correctly_configured
    expect: sbd_startmode == "always"
```
Also https://github.com/trento-project/agent/pull/122 is followed up by using [go-envparse](https://github.com/hashicorp/go-envparse) to parse sbd configuration.

---

Consider also that we have `internal/cluster/sbd.go` doing a similar thing
```go
func getSBDConfig(sbdConfigPath string) (map[string]interface{}, error) {
	sbdConfFile, err := os.Open(sbdConfigPath)
	if err != nil {
		return nil, errors.Wrap(err, "could not open sbd config file")
	}

	defer sbdConfFile.Close()

	sbdConfigRaw, err := io.ReadAll(sbdConfFile)

	if err != nil {
		return nil, errors.Wrap(err, "could not read sbd config file")
	}

	configMap := utils.FindMatches(`(?m)^(\w+)=(\S[^#\s]*)`, sbdConfigRaw)

	return configMap, nil
}
```

This duplication could and should be removed and the loading of an sbd configuration file streamlined for the discovery and for facts gathering.

If that's ok I'd keep that refactoring out of this PR.


internal/factsengine/gatherers/sbd.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/trento-project/agent/internal/factsengine/entities"
+)
+
+const (
+	SBDConfigGathererName = "sbd_config"
+)
+
+// nolint:gochecknoglobals
+var (
+	SbdConfFileError = entities.FactGatheringError{
```
rtorrero, line 19: Just for consistency, I'd either always use `SBDConfig` (as you do in the gatherer name) or `SbdConf` as prefix, but not both.

My personal preference is for the above

rtorrero (approved):
LGTM, just that minor comment about naming

internal/factsengine/gatherers/sbd.go
arbulu89: And I would prefix the error types with `sbd-config` to match

internal/factsengine/gatherers/sbd.go
```diff
@@ -0,0 +1,94 @@
...
+
+func NewSBDGatherer(configFile string) *SBDGatherer {
+	return &SBDGatherer{
+		configFile,
+	}
+}
+
+func (g *SBDGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting SBD Facts gathering")
```
arbulu89, line 51: `SBD config`

internal/factsengine/gatherers/sbd.go
```diff
@@ -0,0 +1,94 @@
...
+	conf, err := loadSbdConf(g.configFile)
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, requestedFact := range factsRequests {
+		var fact entities.Fact
+		if value, found := conf[requestedFact.Argument]; found {
+			fact = entities.NewFactGatheredWithRequest(requestedFact, &entities.FactValueString{Value: value})
```
arbulu89, line 62: Most probably we want to use the `entities.ParseStringToFactValue` function here. The values are not always strings. For example, `SBD_WATCHDOG_TIMEOUT` value is an integer

internal/factsengine/gatherers/sbd_test.go
```diff
@@ -81,26 +65,39 @@ func (suite *SBDGathererTestSuite) TestSBDGatherer() {
 			Gatherer: "sbd_config",
 			Argument: "TEST2",
 		},
+		{
+			Name:     "sbd_unexistent",
+			Gatherer: "sbd_config",
+			Argument: "SBD_THIS_DOES_NOT_EXIST",
```
arbulu89, line 71: We could update one of the entries to test the integer type

test/fixtures/discovery/cluster/sbd/invalid_sbd_config
```diff
@@ -0,0 +1,3 @@
+SBD_DEVICE: /dev/vdc;/dev/vdb
```
arbulu89, line 1: Could we maybe rename the file to `sbd_config_invalid`. 
All the other fixture files start with this prefix

arbulu89 (approved):
It looks good!
I have commented some minor things, but I think they can be changed without any new review
PD: I would change the filename of the gatherer to `sbd_config` maybe. As we will have other sbd gatherers in the future

PR #144, opened 2022-11-18
Corosynccmap ctl gatherer port

rtorrero:
This PR reenables and applies the lastest changes on the corosynccmapctl fact gathering.

A question came up as I was porting this:

- In this gatherer, we issue one `corosynccmap-ctl` call and create a map with the output. All facts that we could look for are (or should be) contained in that map. If the call to `corosynccmap-ctl` fails, we return a `CorosyncCmapCtlCommandError`. In the `package_version` gatherer however, we issue one call per fact request. If we have an 0 we know that most likely, the requested package does not exist, so it makes sense to continue to request other Facts. But if the `rpm` binary itselfs fails to execute, we treat this in the same way (we continue to request other facts regardless the fact that they are not going to be executed). Should we revisit that gatherer and make a disctinction? or do we consider such errors irrelevant for now?


internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,69 @@
...
+	facts := []entities.Fact{}
+	log.Infof("Starting %s facts gathering process", CorosyncCmapCtlGathererName)
+
+	corosyncCmapctl, err := s.executor.Exec(
+		"corosync-cmapctl", "-b")
+	if err != nil {
+		return nil, CorosyncCmapCtlCommandError.Wrap(err.Error())
+	}
+
+	corosyncCmapctlMap := utils.FindMatches(`(?m)^(\S*)\s\(\S*\)\s=\s(.*)$`, corosyncCmapctl)
```
arbulu89, line 52: Maybe the regular expression could be stored as constant

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+
+		if value, ok := corosyncCmapctlMap[factReq.Argument]; ok {
+			fact = entities.NewFactGatheredWithRequest(factReq, &entities.FactValueString{Value: fmt.Sprint(value)})
```
arbulu89, line 58: Use [ParseStringToFactValue](https://github.com/trento-project/agent/blob/main/internal/factsengine/entities/fact_value.go#L91).
We have plenty of integers here

internal/factsengine/gatherers/corosynccmapctl_test.go
```diff
@@ -42,18 +43,53 @@ func (suite *CorosyncCmapctlTestSuite) TestCorosyncCmapctlGathererMissingFact()
...
+		Message: "error while executing corosynccmap-ctl: executable file not found in $PATH",
+		Type:    "corosync-cmapctl-command-error",
+	}
+
+	factResults, err := c.Gather(factRequests)
+
+	suite.EqualError(err, expectedError.Error())
+
+	suite.Empty(factResults)
+
```
arbulu89, line 84: Let's remove this final empty line here

internal/factsengine/gatherers/corosynccmapctl_test.go
```diff
@@ -85,50 +121,29 @@ func (suite *CorosyncCmapctlTestSuite) TestCorosyncCmapctlGatherer() {
...
+	expectedResults := []entities.Fact{
 		{
 			Name:  "quorum_provider",
-			Value: "corosync_votequorum",
+			Value: &entities.FactValueString{Value: "corosync_votequorum"},
 		},
 		{
 			Name:  "totem_max_messages",
-			Value: "20",
+			Value: &entities.FactValueString{Value: "20"},
```
arbulu89, line 131: This should actually be of number type

internal/factsengine/gatherers/gatherer.go
```diff
@@ -10,10 +10,11 @@ type FactGatherer interface {
 
 func StandardGatherers() map[string]FactGatherer {
 	return map[string]FactGatherer{
-		CorosyncConfGathererName:   NewDefaultCorosyncConfGatherer(),
-		HostsFileGathererName:      NewDefaultHostsFileGatherer(),
-		SystemDGathererName:        NewDefaultSystemDGatherer(),
-		PackageVersionGathererName: NewDefaultPackageVersionGatherer(),
-		SBDConfigGathererName:      NewDefaultSBDGatherer(),
+		CorosyncConfGathererName:    NewDefaultCorosyncConfGatherer(),
```
arbulu89, line 13: Could we order these alphabetically?

arbulu89 (changes requested):
Hey @rtorrero ,
Nice!
We at least need to fix the the type usage and use the `ParseString..` function
Green light from my side afterwards

internal/factsengine/gatherers/corosynccmapctl.go
nelsonkopliku: We might want to consider also lists/maps.
For instance we need to determine that _Corosync is running with at least 2 rings_, meaning that in the command output we'd find something like the following (if I understood Check 32CFC6)
```shell
$ corosync-cmapctl -b
...
totem.interface.0.bindnetaddr (str) = 10.80.1.12
...
totem.interface.1.bindnetaddr (str) = 10.80.1.13
...
```
What should we pass as an argument to this gatherer to get the rings and use them in the Spec via `facts.running_corosync_rings.len() == 2`?

This can be very well deferred if too much for this PR, anyway this gatherer already allows to deliver 7 Checks :rocket:

nelsonkopliku (commented):
This is HUGE! 7 extra Checks are supported with this Gatherer.
Great job!

Nothing much to add to what Xabi already commented, except that we'd need to take into account also lists/maps to support all scenarios. Up to you whether to add it now or later.

internal/factsengine/gatherers/corosynccmapctl.go
arbulu89: @nelsonkopliku That's a really fair point.
I didn't consider it myself.
From my side, if it is easy to implement. We could/should do it. It would give the gatherer much more flexibility. Otherwise we are limited to gather ending values.
@rtorrero maybe we can check how easy it is?
I'm not against merging what we have now, but we should track the debt

arbulu89: @CDimonaco maybe you have some idea how we could do this transformation? from dot separated strings to a map

rtorrero:
> This is HUGE! 7 extra Checks are supported with this Gatherer. Great job!
> 
> Nothing much to add to what Xabi already commented, except that we'd need to take into account also lists/maps to support all scenarios. Up to you whether to add it now or later.

It should be easy I think, I'll try to add it. Let's see if future me comes back to bite my ass :smile:

rtorrero:
@nelsonkopliku @arbulu89 I'm making progress on allowing the whole file to be exported as a map, but I'd suggest to get this merged already and get the changes on a separate PR so we allow this gatherer to be used already for cases that don't required the whole map. WDYT?

internal/factsengine/gatherers/corosynccmapctl_test.go
```diff
@@ -8,29 +8,30 @@ import (
...
+	suite.mockExecutor = new(utilsMocks.CommandExecutor)
 }
 
 func (suite *CorosyncCmapctlTestSuite) TestCorosyncCmapctlGathererMissingFact() {
 	mockOutputFile, _ := os.Open(helpers.GetFixturePath("gatherers/corosynccmap-ctl.output"))
 	mockOutput, _ := io.ReadAll(mockOutputFile)
 	suite.mockExecutor.On("Exec", "corosync-cmapctl", "-b").Return(mockOutput, nil)
 
-	c := NewCorosyncCmapctlGatherer(suite.mockExecutor)
+	c := gatherers.NewCorosyncCmapctlGatherer(suite.mockExecutor)
```
arbulu89, line 34: I think we could perfectly add this use case in the "happy path" test, but well, it is fine this way too

arbulu89 (approved):
@rtorrero I'm fine on merging this PR and working right away in the improvement.
I wouldn't put the task as closed and documentation based on the current functionality.

PR #145, opened 2022-11-18
Refactor sbd loading

nelsonkopliku:
## Description

Sbd configuration gets read in the two major processes of the agent:
- Discovery (cluster)
- Facts Gathering (`sbd_config` gatherer)

This PR reduces some duplication of sbd configuration file loading and in discovery/facts gathering.

Notes:
`utils.FindMatches('(?m)^(\w+)=(\S[^#\s]*)', sbdConfigRaw)` has been superseded by `envparse.Parse(sbdConfigFile)`.
A subtle difference here is that `FindMatches` ignores empty values like `SBD_OPTS=` while `envparse` decodes to empty string.

This means that the cluster discovery would also have `"SBD_OPTS": ""` among the other configs.

~Would this break something? :sweat_smile:~ I tried and it does not break things up. 
~Should we make sure to skip empty entries when parsing?~ I believe we should return sbd config entries even if they're empty, see https://github.com/trento-project/agent/pull/145#issuecomment-1321746186

### DIFF
--- a/internal/cluster/sbd.go
+++ b/internal/cluster/sbd.go
@@ -2,12 +2,12 @@ package cluster
 
 import (
 	"fmt"
-	"io"
 	"os"
 	"regexp"
 	"strconv"
 	"strings"
 
+	"github.com/hashicorp/go-envparse"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/pkg/errors"
@@ -25,7 +25,7 @@ const (
 type SBD struct {
 	cluster string
 	Devices []*SBDDevice
-	Config  map[string]interface{}
+	Config  map[string]string
 }
 
 type SBDDevice struct {
@@ -58,10 +58,10 @@ func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath stri
 	var s = SBD{
 		cluster: cluster,
 		Devices: nil, // TODO check me, no slice of pointers needed
-		Config:  map[string]interface{}{},
+		Config:  map[string]string{},
 	}
 
-	c, err := getSBDConfig(sbdConfigPath)
+	c, err := LoadSbdConfig(sbdConfigPath)
 	s.Config = c
 
 	if err != nil {
@@ -70,7 +70,7 @@ func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath stri
 		return s, fmt.Errorf("could not find SBD_DEVICE entry in sbd config file")
 	}
 
-	sbdDevice, ok := c["SBD_DEVICE"].(string)
+	sbdDevice, ok := c["SBD_DEVICE"]
 	if !ok {
 		return s, fmt.Errorf("could not cast sbd device to string, %v", c["SBD_DEVICE"])
 	}
@@ -86,23 +86,25 @@ func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath stri
 	return s, nil
 }
 
-func getSBDConfig(sbdConfigPath string) (map[string]interface{}, error) {
-	sbdConfFile, err := os.Open(sbdConfigPath)
+func LoadSbdConfig(sbdConfigPath string) (map[string]string, error) {
+	sbdConfigFile, err := os.Open(sbdConfigPath)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not open sbd config file")
 	}
 
-	defer sbdConfFile.Close()
+	defer func() {
+		err := sbdConfigFile.Close()
+		if err != nil {
+			log.Error(err)
+		}
+	}()
 
-	sbdConfigRaw, err := io.ReadAll(sbdConfFile)
-
+	conf, err := envparse.Parse(sbdConfigFile)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not read sbd config file")
+		return nil, errors.Wrap(err, "could not parse sbd config file")
 	}
 
-	configMap := utils.FindMatches(`(?m)^(\w+)=(\S[^#\s]*)`, sbdConfigRaw)
-
-	return configMap, nil
+	return conf, nil
 }
 
 func NewSBDDevice(executor utils.CommandExecutor, sbdPath, device string) SBDDevice {
--- a/internal/cluster/sbd_test.go
+++ b/internal/cluster/sbd_test.go
@@ -277,10 +277,11 @@ func (suite *SbdTestSuite) TestLoadDeviceDataError() {
 	suite.EqualError(err, "sbd dump command error: error;sbd list command error: error")
 }
 
-func (suite *SbdTestSuite) TestGetSBDConfig() {
-	sbdConfig, err := getSBDConfig(helpers.GetFixturePath("discovery/cluster/sbd/sbd_config"))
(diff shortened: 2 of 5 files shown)


fabriziosestito (changes requested):
> Would this break something? sweat_smile

@nelsonkopliku  I think that before considering changing this, this question should be answered by adding automated tests where needed

internal/cluster/sbd.go
```diff
@@ -25,7 +25,7 @@ const (
 type SBD struct {
 	cluster string
 	Devices []*SBDDevice
-	Config  map[string]interface{}
+	Config  map[string]string
```
fabriziosestito, line 28: nice!

internal/cluster/sbd.go
```diff
@@ -86,23 +100,31 @@ func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath stri
...
+		}
+	}()
 
+	conf, err := envparse.Parse(sbdConfigFile)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not read sbd config file")
+		return nil, &SBDConfigLoadingError{
+			Type:  SDBConfigDecodingError,
+			inner: errors.Wrap(err, "could not read sbd config file"),
+		}
```
fabriziosestito, line 124: in case we could remove the extra entries so the output is the same to what we had with the regex

nelsonkopliku:
>  adding automated tests where needed

I made this change on trento web https://github.com/trento-project/web/pull/996 and let e2e tests tell us if integration does not break :sweat_smile: 

Would that be enough?

nelsonkopliku:
@fabriziosestito made the change you suggested.
Now if an existing entry in sbd config is an empty string, that entry gets removed from the parsed config.

This also means that if we do something like
```yaml
facts:
  - name: sbd_opts
    gatherer: sbd_config
    argument: SBD_OPTS # or any other possibly empty entry
```
We'll get a `FactGatheringError` saying `sbd-config-value-not-found`.
We can start like this, it's good enough I believe, anyway I see a little discrepancy with the fact that the config value was actually there, it was just empty.
It depends on what meaning we want to give to the emptiness of a config entry and if that state is an error or something we'd like to know as a fact in our expectations.

internal/cluster/sbd.go
```diff
@@ -86,23 +100,37 @@ func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath stri
 	return s, nil
 }
 
-func getSBDConfig(sbdConfigPath string) (map[string]interface{}, error) {
-	sbdConfFile, err := os.Open(sbdConfigPath)
+func LoadSbdConfig(sbdConfigPath string) (map[string]string, error) {
+	sbdConfigFile, err := os.Open(sbdConfigPath)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not open sbd config file")
+		return nil, &SBDConfigLoadingError{
```
arbulu89, line 106: Do we need such a tailored error handling?
Why don't we return how we were returning before?
`errors.Wrap(err, "could not open sbd config file")`
It looks like an overkill to me

internal/factsengine/gatherers/sbd.go
```diff
@@ -72,22 +70,21 @@ func (g *SBDGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.F
 	return facts, nil
 }
 
-func loadSbdConf(sbdConfigPath string) (map[string]string, error) {
-	sbdConfigFile, err := os.Open(sbdConfigPath)
-	if err != nil {
-		return nil, SBDConfigFileError.Wrap(err.Error())
-	}
+func loadSbdConfig(sbdConfigPath string) (map[string]string, error) {
```
arbulu89, line 73: I find this function a bit useless here.
Why don't we simply use `cluster.LoadSbdConfig` up in the code, and a put `SBDConfigDecodingError` together with the error message.
The result would be the same, and we would have less "boilerplate" code

internal/factsengine/gatherers/sbd.go
```diff
@@ -50,7 +48,7 @@ func (g *SBDGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.F
 	facts := []entities.Fact{}
 	log.Infof("Starting SBD config Facts gathering")
 
-	conf, err := loadSbdConf(g.configFile)
+	conf, err := loadSbdConfig(g.configFile)
```
arbulu89, line 51: Here is where we should put the error handling in my opinion.
Directly here, one one error, and that's all

arbulu89 (commented):
Folks, sorry for arriving late, but I think that the removal of empty values is a bad idea. We are hiding real configuration, that might be useful in the checks for example.
Senging a new field with an empty value shouldn't be a problem for the web, as most probably we don't use that field.

Besides that, I find the error handling way complicated

internal/cluster/sbd.go
nelsonkopliku: At the end I returned to my first proposal of not skipping empty values in config and covered in test.

If this is sorted out we need to agree on https://github.com/trento-project/web/pull/996.

internal/cluster/sbd.go
```diff
@@ -86,23 +86,20 @@ func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath stri
...
+func LoadSbdConfig(sbdConfigPath string) (map[string]string, error) {
+	sbdConfigFile, err := os.Open(sbdConfigPath)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not open sbd config file")
 	}
 
-	defer sbdConfFile.Close()
-
-	sbdConfigRaw, err := io.ReadAll(sbdConfFile)
+	defer sbdConfigFile.Close()
```
arbulu89, line 95: We should aim of checking the error output of the `Close` function.
At least to log it

internal/cluster/sbd_test.go
```diff
@@ -278,9 +278,10 @@ func (suite *SbdTestSuite) TestLoadDeviceDataError() {
 }
 
 func (suite *SbdTestSuite) TestGetSBDConfig() {
```
arbulu89, line 280: Let's change the test name to `TestLoadSbdConfig`

arbulu89 (approved):
@nelsonkopliku Green light from my side.
I commented 2 minor things that don't need an extra review.

internal/factsengine/gatherers/sbd_test.go
```diff
@@ -17,6 +17,14 @@ func TestSBDGathererTestSuite(t *testing.T) {
 	suite.Run(t, new(SBDGathererTestSuite))
 }
 
+func (suite *SBDGathererTestSuite) TestDefaultSBDConfigGagherer() {
```
arbulu89, line 20: As agreed offline, let's remove this test. We accept the coverage reduction in this case, as in fact, the reduction simply comes for the fact that there are less lines of code. This function was not tested beforehand neither

fabriziosestito: +1, I lifted the threshold to 1%

nelsonkopliku: Alright, thanks!

internal/cluster/sbd.go
nelsonkopliku: changed

PR #146, opened 2022-11-22
Generic get value

arbulu89:
Make `GetValue` a public function. Accessing to FactValue objects using a `dot` way (`value.other_value.0`, etc) will be a common thing next (for example, crm_mon and cibadmin will use it).
As a side effect, this gives a much more flexible way of testing `GetValue` without depending on real gatherer fixtures.
I have refactored the `fact_value_test.go` tests to use `_test` package and a proper testing suite.

### DIFF
--- a/internal/factsengine/entities/fact_value.go
+++ b/internal/factsengine/entities/fact_value.go
@@ -115,6 +115,9 @@ func ParseStringToFactValue(str string) FactValue {
 // foo.bar.buz access the {"foo": {"bar": {"baz": "value"}}}
 // foo.0.buz access the {"foo": [{"buz": "value"}]}
 func GetValue(fact FactValue, values string) (FactValue, *FactGatheringError) {
+	// splitDotAccess returns and empty list if the coming argument is an empty string.
+	// It is used to replace strings.Split as this 2nd returns a one element list with
+	// and empty string in the same scenario
 	splitDotAccess := func(c rune) bool {
 		return c == '.'
 	}
--- a/internal/factsengine/entities/fact_value.go
+++ b/internal/factsengine/entities/fact_value.go
@@ -1,10 +1,20 @@
 package entities
 
 import (
+	"fmt"
 	"strconv"
+	"strings"
 )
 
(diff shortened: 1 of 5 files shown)


internal/factsengine/entities/fact_value.go
```diff
@@ -99,3 +109,44 @@ func ParseStringToFactValue(str string) FactValue {
 
 	return &FactValueString{Value: str}
 }
+
+// GetValue returns a value using a dot access key format from a FactValue.
+// Examples:
+// foo.bar.buz access the {"foo": {"bar": {"baz": "value"}}}
+// foo.0.buz access the {"foo": [{"buz": "value"}]}
+func GetValue(fact FactValue, values string) (FactValue, *FactGatheringError) {
```
fabriziosestito, line 117: wouldn't be possible to expose this as method of `FactValueMap` and `FactValueList` so that in the corosync conf gahterer we just can call the method without changing the type to the interface in the `corosyncConfToMap` function?

CDimonaco: Agree with @fabriziosestito this should be a method on the structs

arbulu89: mmm, let me check the option. I guess it should be totally doable.
PD: Fabriozio asking to move from "pure" functions to OOP kind of functions. Not sure if this was not written by copilot... XD

CDimonaco: The gopher daemon corrupts everyone, a "rustichella bite" after another 🤣

arbulu89: Done!
I have attached the function only to `FactValueMap` even though it could be useful for lists as well, but we don't have this case yet, so we can think of adding it once we need it

fabriziosestito: I can argue this is stil a pure function!

CDimonaco (approved):
lgtm good job

nelsonkopliku (approved):

PR #147, opened 2022-11-24
Restructure project folders structure

arbulu89:
Refactor folder structure. At the end, we still keep the major part of the code in `internal` and only publish the data needed to implemente gatherer plugins.
I have removed some unused functions and move other closer to their call, to reduce the "trash" utils file


CDimonaco (approved):
Perfect, LGTM

PR #154, opened 2022-12-14
Retrieve agent id command

nelsonkopliku:
This PR adds a command to reliably retrieve the agent id from a trento-agent installation.

Here's the usage
```bash
$ trento-agent id
f6fe2f71-42cc-482f-877f-00d9b4592d83
```

And here's  a sample output in case of an error
```bash
$ trento-agent id
FATA[0000] Failed to get Agent id: open /etc/machine-idz: no such file or directory
```

### DIFF
--- a/cmd/id.go
+++ b/cmd/id.go
@@ -1,0 +1,26 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/trento-project/agent/internal/agent"
+)
+
+func NewAgentIDCmd() *cobra.Command {
+	idCmd := &cobra.Command{ //nolint
+		Use:   "id",
+		Short: "Print the agent identifier",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			agentID, err := agent.GetAgentID()
+			if err != nil {
+				return err
+			}
+			_, err = os.Stdout.WriteString(agentID)
+
+			return err
+		},
+	}
+
+	return idCmd
+}
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -39,6 +39,7 @@ that can help you deploy, provision and operate infrastructure for SAP Applicati
 		}
 	})
 
+	rootCmd.AddCommand(NewAgentIDCmd())
 	rootCmd.AddCommand(NewStartCmd())
 	rootCmd.AddCommand(NewFactsCmd())
 	rootCmd.AddCommand(NewVersionCmd())
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -42,7 +42,7 @@ type Config struct {
 
 // NewAgent returns a new instance of Agent with the given configuration
 func NewAgent(config *Config) (*Agent, error) {
-	agentID, err := getAgentID()
+	agentID, err := GetAgentID()
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get the agent ID")
 	}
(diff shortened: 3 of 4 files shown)


cmd/id.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/trento-project/agent/internal/agent"
+)
+
+func NewIDCmd() *cobra.Command {
+	idCmd := &cobra.Command{ //nolint
+		Use:   "id",
+		Short: "Print the Agent identifier",
+		Long:  "The unique identifier for the current trento agent installation",
```
dottorblaster, line 16:
```suggestion
```

dottorblaster: I think we don't need this :grimacing:

dottorblaster (changes requested):
Just a tiny comment then we can merge

cmd/id.go
nelsonkopliku: Removed :wink:

dottorblaster (approved):
LGTM

cmd/id.go
```diff
@@ -0,0 +1,26 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"github.com/spf13/cobra"
+	"github.com/trento-project/agent/internal/agent"
+)
+
+func NewIDCmd() *cobra.Command {
+	idCmd := &cobra.Command{ //nolint
+		Use:   "id",
+		Short: "Print the Agent identifier",
```
arbulu89, line 15: We have the agent word already in small case in other help strings. I would continue using that way

cmd/id.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"fmt"
+
+	log "github.com/sirupsen/logrus"
+
+	"github.com/spf13/cobra"
+	"github.com/trento-project/agent/internal/agent"
+)
+
+func NewIDCmd() *cobra.Command {
```
arbulu89, line 12: Maybe not better `NewAgentIDCmd`?
To follow along the `GetAgetnID` function name

arbulu89 (approved):
I would remove the added documentation.
Besides that, good to go

README.md
```diff
@@ -199,6 +200,26 @@ Examples:
 
 `server-ip` -> `TRENTO_SERVER_IP=https://localhost ./trento-agent start`
 
+## Commands
```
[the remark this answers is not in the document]
nelsonkopliku: Totally agree that there is the risk that it could get outdated very soon.
Indeed my intent is to just give a little reference, not to document everything, keeping the change rate of the doc at minimum :sweat_smile:  

I took inspiration from [datadog agent documentation](https://github.com/DataDog/documentation/blob/master/content/en/agent/guide/agent-commands.md) and looked reasonable. 
No problems in removing it, let's hear from the others as well.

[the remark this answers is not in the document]
fabriziosestito: +1 removing this, we will forget about updating it for sure.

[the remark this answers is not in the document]
nelsonkopliku: gone :wink:

arbulu89 (approved):
Thanks @nelsonkopliku ,
Perfect now from my side

cmd/id.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+func NewAgentIDCmd() *cobra.Command {
+	idCmd := &cobra.Command{ //nolint
+		Use:   "id",
+		Short: "Print the agent identifier",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			agentID, err := agent.GetAgentID()
+			if err != nil {
+				return err
```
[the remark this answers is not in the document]
nelsonkopliku: @CDimonaco are we good with the changes? Anything else is needed prior merging?

PR #156, opened 2022-12-15
Sbd dump gatherer

nelsonkopliku:
Follows up https://github.com/trento-project/agent/pull/74 and enables porting of checks
- B089BE SBD watchdog timeout is set to expected value
- 68626E SBD `msgwait` timeout value is two times the watchdog timeout

A little sanitization of the keys has been applied to simplify usage of the result at expression level (ie `Timeout (watchdog)` becomes `timeout_watchdog`)

```yaml
facts:
  - name: sbd_devices_dump
    gatherer: sbd_dump

expectations:
  # B089BE
  - name: sbd_watchdog_timeout_is_correct
    expect: facts.sbd_devices_dump.values().all(|device| device.timeout_watchdog == values.expected_watchdog_timeout)

  # 68626E
  - name: sbd_msgwait_timeout_is_double_of_watchdog_timeout
    expect: facts.sbd_devices_dump.values().all(|device| device.timeout_msgwait == 2 * device.timeout_watchdog)
```

Doc added here https://github.com/trento-project/wanda/pull/137

### DIFF
--- a/internal/factsengine/gatherers/sbddump.go
+++ b/internal/factsengine/gatherers/sbddump.go
@@ -1,0 +1,147 @@
+package gatherers
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/agent/pkg/factsengine/entities"
+	"github.com/trento-project/agent/pkg/utils"
+)
+
+const (
+	SBDDumpGathererName = "sbd_dump"
+)
+
+// nolint:gochecknoglobals
+var (
+	SBDDevicesLoadingError = entities.FactGatheringError{
+		Type:    "sbd-devices-loading-error",
+		Message: "error loading the configured sbd devices",
+	}
+
+	SBDDumpCommandError = entities.FactGatheringError{
+		Type:    "sbd-dump-command-error",
+		Message: "error while executing sbd dump",
+	}
+)
+
+type SBDDumpGatherer struct {
+	executor    utils.CommandExecutor
+	sbdGatherer *SBDGatherer
+}
+
+func NewDefaultSBDDumpGatherer() *SBDDumpGatherer {
+	return NewSBDDumpGatherer(utils.Executor{}, NewDefaultSBDGatherer())
+}
+
+func NewSBDDumpGatherer(executor utils.CommandExecutor, sbdGatherer *SBDGatherer) *SBDDumpGatherer {
+	return &SBDDumpGatherer{
+		executor:    executor,
+		sbdGatherer: sbdGatherer,
+	}
+}
+
+func (gatherer *SBDDumpGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting %s facts gathering process", SBDDumpGathererName)
+
+	configuredDevices, err := loadDevices(gatherer.sbdGatherer)
+
+	if err != nil {
+		return facts, SBDDevicesLoadingError.Wrap(err.Error())
+	}
+
+	for _, factRequest := range factsRequests {
+		var fact entities.Fact
+
+		if devicesDumps, err := getSBDDevicesDumps(gatherer.executor, configuredDevices); err == nil {
+			fact = entities.NewFactGatheredWithRequest(factRequest, &entities.FactValueMap{Value: devicesDumps})
+		} else {
+			gatheringError := SBDDumpCommandError.Wrap(err.Error())
+
+			fact = entities.NewFactGatheredWithError(factRequest, gatheringError)
+		}
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested %s facts gathered", SBDDumpGathererName)
+	return facts, nil
+}
+
+func loadDevices(gatherer *SBDGatherer) ([]string, error) {
+	sbdDevicesRequest := []entities.FactRequest{
+		{
+			Name:     "configured_devices",
+			Gatherer: "sbd_config",
+			Argument: "SBD_DEVICE",
+		},
+	}
+
+	sbdDevicesFacts, err := gatherer.Gather(sbdDevicesRequest)
+
+	if err != nil {
+		return []string{}, err
+	}
+
+	sbdDevicesFact := sbdDevicesFacts[0]
+
+	if sbdDevicesFact.Error != nil {
+		return []string{}, sbdDevicesFact.Error
+	}
+
+	deviceAsStringValue, ok := sbdDevicesFact.Value.(*entities.FactValueString)
+
+	if !ok {
+		return []string{}, fmt.Errorf("Unable to determine the device name: %s", sbdDevicesFact.Value)
+	}
+
+	return strings.Split(deviceAsStringValue.Value, ";"), nil
+}
+
+func getSBDDevicesDumps(
+	executor utils.CommandExecutor,
+	configuredDevices []string) (map[string]entities.FactValue, error) {
+	var devicesDumps = make(map[string]entities.FactValue)
+
+	for _, device := range configuredDevices {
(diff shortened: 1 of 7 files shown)


rtorrero (commented):
Hey @nelsonkopliku ! thanks for this, after some thought I think you took the right path by changing the gatherer to use data also from the SBDconfig (instead of the mess that it would have been to implement the dependency as I originally suggested)

Btw, is the output right?

I did:
```
^^/W/t/agent >>> ./trento-agent facts gather --gatherer sbd_dump                                                                                                                                                                                   ^(sbd-dump-gatherer+154) 15:06:06 
INFO[2022-12-19 15:06:31] Using config file: /home/rtorrero/.config/trento/agent.yaml 
INFO[2022-12-19 15:06:31] loading plugins                              
INFO[2022-12-19 15:06:31] Starting sbd_dump facts gathering process    
INFO[2022-12-19 15:06:31] Starting SBD config Facts gathering          
INFO[2022-12-19 15:06:31] Requested sbd_dump facts gathered            
INFO[2022-12-19 15:06:31] Gathered fact for "sbd_dump" with argument "": 
INFO[2022-12-19 15:06:31] {
  "Name": "",
  "CheckID": "",
  "Value": {
    "Value": {
      "/dev/vdb": {
      ...
```

Why do we have a nested `Value`?

Apart from that, consider in a future PR adding support for retrieving specific keys using arguments (rather than the whole map). We have done this in other gatherers and I think it might be a good standard to adopt for gatherers in general.

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,147 @@
...
+	for key, value := range SBDDumpMap {
+		valueAsString, ok := value.(string)
+
+		if !ok {
+			err := fmt.Errorf(`Unable to parse sbd dump entry "%s" as string: %s`, key, value)
+			log.Error(err)
+			return nil, err
+		}
+
+		key = regexp.MustCompile(`[()]`).ReplaceAllString(key, "")
```
arbulu89, line 141: We should compile this regex expression outside, in some variable. Or at least outside of the for loop.

arbulu89 (changes requested):
I find the usage of other gatherers dependency a big smell.
I would get the devices in a plain way, using the `cluster.LoadSbdConfig` function and getting the key from it.
Less dependency and leaner code.
Besides that, some minor nitpick

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested %s facts gathered", SBDDumpGathererName)
+	return facts, nil
+}
+
+func loadDevices(gatherer *SBDGatherer) ([]string, error) {
+	sbdDevicesRequest := []entities.FactRequest{
```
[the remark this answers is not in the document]
nelsonkopliku: Guess that's wiser indeed.

Initially gatherer composition sounded interesting, to avoid duplicating some the code from another gatherer, yet I realize that in this case _a little duplication is better than a little dependency_ :sweat_smile: .

Changing to `cluster.LoadSbdConfig`

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,147 @@
...
+		{
+			Name:     "configured_devices",
+			Gatherer: "sbd_config",
+			Argument: "SBD_DEVICE",
```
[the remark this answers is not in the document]
nelsonkopliku: Didn't consider that :eyes: thanks for bringing up.
I'll se what heppens when adjusting to `cluster.LoadSbdConfig` instead of gatherer composition.

In any case the possibilities are two:
- `SBD_DEVICE` is not even present, which results in a gathering error
- `SBD_DEVICE` is empty which results in an empty list of devices
which of these determines diskless sbd?

rtorrero:
> @rtorrero thanks for the review.
> 
> > Why do we have a nested `Value`?
> 
> That's the internal representation of different types of values that can be returned as facts, and that's suppported by a bit of polymorphism here https://github.com/trento-project/agent/blob/main/pkg/factsengine/entities/fact_value.go

true, we already discussed this, vacations finished frying the last working part of my brain :smile:

internal/factsengine/gatherers/sbddump.go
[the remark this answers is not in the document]
nelsonkopliku: I went cautious with the [latest changes](https://github.com/trento-project/agent/pull/156/files#diff-d78d11a990a55f93c99ffa880338db0b0e4154502b8c802cbd6620d1b987a7e7R84) in `loadDevices`: if the entry `SBD_DEVICE` is not in the config or it's empty an error is returned, excluding the possibility to use the `sbd_dump` gatherer with diskless sbd.

Would this be enough?

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (gatherer *SBDDumpGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting %s facts gathering process", SBDDumpGathererName)
+
+	configuredDevices, err := loadDevices(gatherer.sbdConfigFile)
+
+	if err != nil {
+		return facts, SBDDevicesLoadingError.Wrap(err.Error())
```
arbulu89, line 57: The normal convention would be to return `nil, error` here

internal/factsengine/gatherers/sbddump.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func loadDevices(sbdConfigFile string) ([]string, error) {
+	conf, err := cluster.LoadSbdConfig(sbdConfigFile)
+
+	if err != nil {
+		return nil, err
+	}
+
+	configuredDevices, found := conf["SBD_DEVICE"]
```
arbulu89, line 84: One things here. The SBD_DEVICE value can come between quotes:
```
SBD_DEVICE="/dev/vdc;/dev/vdb"
```
Are we handling this properly. I know we do in the main sbd code code, but I don't know if the `LoadSbdConfig` does the magic.
We could include a test case for this

internal/factsengine/gatherers/sbddump_test.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	gatheredFacts, err := sbdDumpGatherer.Gather(factRequests)
+
+	expectedFacts := []entities.Fact{
+		{
+			Name:  "sbd_devices_dump",
+			Value: nil,
+			Error: &entities.FactGatheringError{
+				Message: "error while executing sbd dump: a failure",
```
arbulu89, line 80: I think here the desired output would to show which device failed. So maybe we could update the wrapped or the returned error somehow

arbulu89 (commented):
Hey @nelsonkopliku ,
Thank you for the changes. The code looks much better now in my opinion.
Have a look on the quoted thing in the `SBD_DEVICE` I commented. If that works, we are ready to go

internal/factsengine/gatherers/sbddump.go
nelsonkopliku: I tested manually changing the fixture file to both with/without quotes and works.

If you think the extra test is really needed, it could be a candidate for another PR :sweat_smile:

arbulu89: Why other PR? Come on man! XD

nelsonkopliku: :stuck_out_tongue_closed_eyes: ok

arbulu89 (approved):
@nelsonkopliku 
I was not expecting the test this way. I would expect a new test in the gatherer itself, as using the `LoadSbdConfig` funtion is a internal tech detail.
Anyway, I see that it works, so feel free to merge if you want.

PR #157, opened 2022-12-19
Fix list conversion issues in the xml gatherer

arbulu89:
During the last sprint review I noticed that in some corner cases, the conversion to list was doing something strange.
At the end I have been forced to change how it was, adding some recursive code to convert the desired fields to list properly.

The code might looks complex, what it is what is is in this sort of things on golang...

### DIFF
--- a/internal/factsengine/gatherers/cibadmin_test.go
+++ b/internal/factsengine/gatherers/cibadmin_test.go
@@ -101,6 +101,12 @@ func (suite *CibAdminTestSuite) TestCibAdminGather() {
 			Argument: "cib.not_found.crm_config",
 			CheckID:  "check3",
 		},
+		{
+			Name:     "primitives",
+			Gatherer: "cibadmin",
+			Argument: "cib.configuration.resources.primitive.0",
+			CheckID:  "check4",
+		},
 	}
 
 	factResults, err := p.Gather(factRequests)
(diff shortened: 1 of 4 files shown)


CDimonaco:
I understand the complexity of the task and golang is really verbose, could we add some comments through the function? Just to have a better understanding of whats'happening? Due to all the casting I don't think we can achieve a really better refactoring, so the comments imho are the best way to address this right now :D

internal/factsengine/gatherers/cibadmin_test.go
```diff
@@ -101,6 +101,12 @@ func (suite *CibAdminTestSuite) TestCibAdminGather() {
 			Argument: "cib.not_found.crm_config",
 			CheckID:  "check3",
 		},
+		{
+			Name:     "primitives",
+			Gatherer: "cibadmin",
+			Argument: "cib.configuration.resources.primitive.0",
```
nelsonkopliku, line 107: I would have added also a case for `cib.configuration.resources.primitive` which returns a list, no blocker tho for me.

nelsonkopliku (approved):
Did you say `interface{}`? :smile: 

LGTM. If some comment help, as @CDimonaco is suggesting, it'd be ok for me.

CDimonaco (changes requested):
Brackets on the case and comments and we are good to go, nothing on the code itself but only on readability and future maintenance

internal/factsengine/gatherers/cibadmin_test.go
arbulu89: @nelsonkopliku A list scenario is already covered within this argument value.
I chose specifically this primitive resource, as it has a single `nvpair`, which is in the `elementsToConvert` list. With that, I already cover that this unique element is converted to a list.

internal/factsengine/gatherers/xml.go
```diff
@@ -38,26 +32,31 @@ func parseXMLToFactValueMap(xmlContent []byte, elementsToList []string) (*entiti
...
-		values := map[string]interface{}{
-			key: value,
+func convertListElements(currentMap map[string]interface{}, elementsToList map[string]bool) map[string]interface{} {
+	convertedMap := make(map[string]interface{})
+	for key, value := range currentMap {
+		switch assertedValue := value.(type) {
+		case map[string]interface{}:
+			convertedMap[key] = convertListElements(assertedValue, elementsToList)
+		case []interface{}:
+			newList := []interface{}{}
```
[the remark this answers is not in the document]
arbulu89: Updated: https://github.com/trento-project/agent/pull/157/commits/0e0e87dcc5a2f519ac348cd9d8235e6f7d6eaf35

internal/factsengine/gatherers/cibadmin_test.go
nelsonkopliku: All good then, thanks!

CDimonaco (approved):

PR #160, opened 2022-12-22
Hidden agent id flag

arbulu89:
Add a new hidden `agent-id` flag to have the option to fake the agent id.
This comes handy during development/testing, as you can have predefined IDs, for whatever need you have.

I have added some dependency injection with Afero to manage the filesystem in the `GetAgentID` function. Not so nice to be passing this value all over the places, but well, it makes testing way better.

PD: Finally, I have changed the horrible testing of the `config_test.go` file using the teardown :facepalm:

### DIFF
--- a/cmd/id.go
+++ b/cmd/id.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"os"
 
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/trento-project/agent/internal/agent"
 )
@@ -12,7 +13,7 @@ func NewAgentIDCmd() *cobra.Command {
 		Use:   "id",
 		Short: "Print the agent identifier",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			agentID, err := agent.GetAgentID()
+			agentID, err := agent.GetAgentID(afero.NewOsFs())
 			if err != nil {
 				return err
 			}
--- a/test/helpers/filesystem.go
+++ b/test/helpers/filesystem.go
@@ -1,0 +1,25 @@
+package helpers
+
+import (
+	"github.com/spf13/afero"
+)
+
+const (
+	DummyMachineID = "dummy-machine-id"
+	DummyAgentID   = "779cdd70-e9e2-58ca-b18a-bf3eb3f71244"
+
+	machineIDPath = "/etc/machine-id"
+)
+
+// MockMachineIDFile mocks the /etc/machine-id file to have a known value
+func MockMachineIDFile() afero.Fs {
+	fileSystem := afero.NewMemMapFs()
+
+	err := afero.WriteFile(fileSystem, machineIDPath, []byte(DummyMachineID), 0644)
+
+	if err != nil {
+		panic(err)
+	}
+
+	return fileSystem
+}
--- a/README.md
+++ b/README.md
@@ -224,12 +224,20 @@ You can install it with `go install github.com/vektra/mockery/v2`.
 
 > Be sure to add the `mockery` binary to your `$PATH` environment variable so that `make` can find it. That usually comes with configuring `$GOPATH`, `$GOBIN`, and adding the latter to your `$PATH`.
 
-
-
 > Please note that the `trento agent` component requires to be running on
 > the OS (_not_ inside a container) so, while it is technically possible to run `trento agent`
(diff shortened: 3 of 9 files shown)


dottorblaster (changes requested):
Looks good, the only thing: maybe having something more explicit like `--force-agent-id`?

test/helpers/filesystem.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"github.com/spf13/afero"
+)
+
+const (
+	DummyMachineID = "dummy-machine-id"
+	DummyAgentID   = "779cdd70-e9e2-58ca-b18a-bf3eb3f71244"
+
+	machineIDPath = "/etc/machine-id"
```
nelsonkopliku, line 11: This declaration is being duplicated. Would it make sense to make it public from `agent` package and use it here?

dottorblaster (approved):
LGTM

CDimonaco (approved):
LGTM

cmd/id.go
```diff
@@ -12,7 +13,7 @@ func NewAgentIDCmd() *cobra.Command {
 		Use:   "id",
 		Short: "Print the agent identifier",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			agentID, err := agent.GetAgentID()
+			agentID, err := agent.GetAgentID(afero.NewOsFs())
```
nelsonkopliku, line 16: This PR opens the need to refactor also `id` command I guess, because as of now it will always get the _real agent id_.

What I would expect is that when `./trento-agent id` is executed it returns either what has been configured in agent's config file or the real agent id.

As an addition I wouldn't expect `./trento-agent id --force-agent-id=blaa`, as it would be fancy :sweat_smile: 

Finally we'd need to take care of having the same defaults for both the commands, otherwise in one case viper would complain about not having some expected flags.

test/helpers/filesystem.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// MockMachineIDFile mocks the /etc/machine-id file to have a known value
+func MockMachineIDFile() afero.Fs {
```
nelsonkopliku, line 15: I find this a bit misleading as we're mocking the whole filesystem, not only the file :sweat_smile:

nelsonkopliku (commented):
Looks good, even if I am not a fan either of passing afero around, however, that's what we have :sweat_smile: 

I find more compelling the changes that this PR requires to make, at least for me, to the `id` command.

test/helpers/filesystem.go
arbulu89: I wouldn't. This is an internal concept. Better a small duplication here in my opinion

cmd/id.go
arbulu89: I wouldn't go in this direction. This `force-agent-id` is a hidden flag for development/test purposes, and it shouldn't be used together with the `id` command, which is used to get the real id. It is only for the `start` function, which is the one that it makes sense.
Using `./trento-agent id --force-agent-id=blaa` doesn't make any sense.

test/helpers/filesystem.go
nelsonkopliku: ok

test/helpers/filesystem.go
arbulu89: How would you call it?
I find the name together with the docstring pretty meaningful, but I wrote msyelf the function, so I'm biased

cmd/id.go
nelsonkopliku: Yeah true, that's just for test/dev so if that's how it has to be I can live with this little inconsistency :wink:

> Using ./trento-agent id --force-agent-id=blaa doesn't make any sense.
>
Yes, we're saying the same about this :smile:

test/helpers/filesystem.go
nelsonkopliku: I wouldn't call it :stuck_out_tongue_winking_eye: 

However, I can think of `MockFilesystemWithMachineID`, `MockFilesystemMachineID`, which are by no means any better than `MockMachineIDFile` at the end :sweat_smile: 

So up to you.

rtorrero (approved):
All looks good to me, but maybe considering issuing a warning when `force-agent-id` is used.

PR #164, opened 2022-12-30
Add initial support to verify the password for the hacluster user

rtorrero:
I've updated the initial attempt to include the latest changes. I've modified it a bit for security reasons so that only whitelisted users might be tested and only against hardcoded passwords to avoid attempts to abuse this to gain the password of a certain user.


internal/factsengine/gatherers/verifypassword.go
```diff
@@ -9,45 +9,52 @@ import (
 	crypt "github.com/tredoe/osutil/user/crypt"
 	sha512crypt "github.com/tredoe/osutil/user/crypt/sha512_crypt"
 	"github.com/trento-project/agent/pkg/factsengine/entities"
+	"github.com/trento-project/agent/pkg/utils"
 )
 
 const (
 	VerifyPasswordGathererName = "verify_password"
+	checkableUsernames         = "hacluster"
```
nelsonkopliku, line 17: This could be a map of username => wellKnownPassword 
```go
var wellKnownUserPassowrdPairs = map[string]string{
	"hacluster": "linux",
}
```
It would allow to easily add more pairs.

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -9,45 +9,52 @@ import (
...
+)
+
+// nolint:gochecknoglobals
+var (
+	VerifyPasswordInvalidUsername = entities.FactGatheringError{
+		Type:    "verify-password-invalid-username",
+		Message: "unknown username or not allowed to check",
```
nelsonkopliku, line 24: We could narrow down to just the _not allowed_ case.

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -9,45 +9,52 @@ import (
...
 		executor,
 	}
 }
 
 /*
-This gatherer expects the next format for the argument: "username:password"
-Where:
-- username: The user which the password is verified
-- password: The password to verify
+This gatherer expects the only the user which password is verified
```
nelsonkopliku, line 43: typo in `...expects the only the user....`

nelsonkopliku (changes requested):
Based on the feedbacks shared during the review we could make some adjustments to the whitelisted username/password pairs.

internal/factsengine/gatherers/verifypassword.go
rtorrero: This would force us to duplicate common passwords for every user we want to check; rather than a map I'd have a separate array of passwords, wdyt?

nelsonkopliku: If I get it right, the duplication that is concerning you is the following
```go
var wellKnownUserPassowrdPairs = map[string]string{
	"hacluster":      "linux",
	"anotheruser":    "linux",
	"yetanotheruser": "linux",
}
```
I could live with that for the time being, yet I see your point.
Feel free to come with changes that address review's feedbacks and the hardcoded password duplication concern.

If it becomes too much of a hassle, consider deferring to a later stage the complexity of linking a list/map of usernames to a list/map of passwords.

rtorrero: I proposed something slightly different in the code, take a look and let me know what you think :+1:

nelsonkopliku: I'd find just more natural working with lists/maps for this usecase, rather than strings that we need to split or look into.
Let's get some feedbacks from others as well.

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -0,0 +1,94 @@
...
+	crypt "github.com/tredoe/osutil/user/crypt"
+	sha512crypt "github.com/tredoe/osutil/user/crypt/sha512_crypt"
+	"github.com/trento-project/agent/pkg/factsengine/entities"
+	"github.com/trento-project/agent/pkg/utils"
+)
+
+const (
+	VerifyPasswordGathererName = "verify_password"
+	checkableUsernames         = "hacluster"
+	unsafePasswords            = "linux"
```
fabriziosestito, line 18: why are we splitting by comma instead of using a slice here?

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -0,0 +1,94 @@
...
+
+		salt, hash, err := g.getSalt(username)
+		if err != nil {
+			log.Error(err)
+		}
+		log.Debugf("Obtained salt using user %s: %s", username, salt)
+
+		crypter := sha512crypt.New()
+		isPasswordWeak := false
+		for _, password := range strings.Split(unsafePasswords, ",") {
```
fabriziosestito, line 67: why don't we use a slice?

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -0,0 +1,94 @@
...
+
+/*
+This gatherer expects only the username for which the password will be verified
+*/
+func (g *VerifyPasswordGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting password verifying facts gathering process")
+
+	for _, factReq := range factsRequests {
+		if !strings.Contains(checkableUsernames, factReq.Argument) {
```
fabriziosestito, line 51: this is prone to error, the string "ha" is contained in the string "hacluster,something"
why don't we use a slice, and check for the existence in the array?

fabriziosestito (changes requested):

internal/factsengine/gatherers/verifypassword.go
fabriziosestito: @nelsonkopliku it depends on the usage. if we agree that we check the same set of passwords for all the users, then a map is not the right solution.

internal/factsengine/gatherers/verifypassword.go
rtorrero: I was trying to avoid not using a const but you are totally right with your comment below, changing it.

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -0,0 +1,98 @@
...
+			log.Error(err)
+		}
+		log.Debugf("Obtained salt using user %s: %s", username, salt)
+
+		crypter := sha512crypt.New()
+		isPasswordWeak := false
+		for _, password := range unsafePasswords {
+			passwordBytes := []byte(password)
+			match := crypter.Verify(hash, passwordBytes)
+			if !errors.Is(match, crypt.ErrKeyMismatch) {
```
nelsonkopliku, line 74: maybe an edge case but `match := crypter.Verify(hash, passwordBytes)` could return a different error than `crypt.ErrKeyMismatch`, without knowing whether the hashes are matching or not ([see here](https://github.com/tredoe/osutil/blob/master/v2/userutil/crypt/sha512_crypt/sha512_crypt.go#L221)) resulting in the detection of a weak password, which would not be necessarily true.
An option could be to handle this case explicitly and return a `FactGatheringError`. Makes sense?

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -0,0 +1,115 @@
...
+		for _, password := range unsafePasswords {
+			passwordBytes := []byte(password)
+
+			matchErr := crypter.Verify(hash, passwordBytes)
+			if matchErr == nil {
+				isPasswordWeak = true
+				break
+			}
+			if matchErr.Error() != crypt.ErrKeyMismatch.Error() {
+				gatheringError = VerifyPasswordCryptError.Wrap(factReq.Argument)
```
nelsonkopliku, line 86: we could use the defined `username`. Up to you.

pkg/utils/utils.go
```diff
@@ -30,3 +30,13 @@ func FindMatches(pattern string, text []byte) map[string]interface{} {
 	}
 	return configMap
 }
+
+// Check if a string is present in a slice of strings
+func StringInSlice(a string, list []string) bool {
```
fabriziosestito, line 35: this is a good usecase for generics, so you can pass any type

```go
func Contains[T comparable](s []T, e T) bool {
    for _, v := range s {
        if v == e {
            return true
        }
    }
    return false
}
```

fabriziosestito (changes requested):

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -0,0 +1,115 @@
...
+		var gatheringError *entities.FactGatheringError = nil
+		for _, password := range unsafePasswords {
+			passwordBytes := []byte(password)
+
+			matchErr := crypter.Verify(hash, passwordBytes)
+			if matchErr == nil {
+				isPasswordWeak = true
+				break
+			}
+			if matchErr.Error() != crypt.ErrKeyMismatch.Error() {
```
nelsonkopliku, line 85: I would have kept the previous `if !errors.Is(matchErr, crypt.ErrKeyMismatch) {`. Up to you.

nelsonkopliku (approved):
LGTM as long as you sort out the generalization mentioned by Fabrizio.

pkg/utils/utils.go
```diff
@@ -31,10 +31,10 @@ func FindMatches(pattern string, text []byte) map[string]interface{} {
 	return configMap
 }
 
-// Check if a string is present in a slice of strings
-func StringInSlice(a string, list []string) bool {
-	for _, b := range list {
-		if b == a {
+// Check if a slice contains a given element
```
fabriziosestito, line 34: starting comments with the function name is interpreted as a docstring:

```go
// Contains checks if a slice contains a given element
```

fabriziosestito (approved):
LGTM Just a tiny comment but feel free to merge it once it is addressed

pkg/utils/utils.go
```diff
@@ -30,3 +30,12 @@ func FindMatches(pattern string, text []byte) map[string]interface{} {
 	}
 	return configMap
 }
+
+func Contains[T comparable](s []T, e T) bool {
```
arbulu89, line 34: We are back again with the contains things XD
We started using the `map[string]bool` to solve this, as it was much more proficient in golang languange, but so be it hehe

arbulu89 (commented):
Late to the party.
I already followed the previous discussions, and I still see this generic `unsafePasswords` list counterintuitive.
I don't think this gatherer will evolve to anything else (I would even hardcode the gatherer only for the `hacluster` user) but if it does, and some user password list must be different, the code will make little sense hehe

PR #168, opened 2023-01-11
Make `corosynccmapctl` gatherer output a map structure

jamie-suse:
Adds the ability for the `corosynccmapctl` gatherer to output a map structure.

e.g. given the following output from `corosync-cmapctl`:
```
nodelist.local_node_pos (u32) = 1
nodelist.node.0.nodeid (u32) = 1
nodelist.node.0.ring0_addr (str) = 10.80.1.11
nodelist.node.1.nodeid (u32) = 2
nodelist.node.1.ring0_addr (str) = 10.80.1.12
```

If a `FactRequest` with `Argument` = `"nodelist.node"` is requested, then the following structure is returned:
```go
{
	Name: "nodes",
	Value: &entities.FactValueMap{
		Value: map[string]entities.FactValue{
			"0": &entities.FactValueMap{
				Value: map[string]entities.FactValue{
					"nodeid":     &entities.FactValueInt{Value: 1},
					"ring0_addr": &entities.FactValueString{Value: "10.80.1.11"},
				},
			},
			"1": &entities.FactValueMap{
				Value: map[string]entities.FactValue{
					"nodeid":     &entities.FactValueInt{Value: 2},
					"ring0_addr": &entities.FactValueString{Value: "10.80.1.12"},
				},
			},
		},
	},
}
```

If a `FactRequest` with `Argument` = a leaf of the output tree e.g. `nodelist.local_node_pos`, then the value is returned:

```go
{
	Name:  "local_node_pos",
	Value: &entities.FactValueInt{Value: 1},
}
```

### DIFF
--- a/internal/factsengine/gatherers/corosynccmapctl.go
+++ b/internal/factsengine/gatherers/corosynccmapctl.go
@@ -9,8 +9,7 @@ import (
 )
 
 const (
-	CorosyncCmapCtlGathererName  = "corosync-cmapctl"
-	corosyncCmapCtlparsingRegexp = `(?m)^(\S*)\s\(\S*\)\s=\s(.*)$`
+	CorosyncCmapCtlGathererName = "corosync-cmapctl"
 )
 
 // nolint:gochecknoglobals
@@ -70,7 +69,7 @@ func corosyncCmapctlOutputToMap(corosyncCmapctlOutput string) *entities.FactValu
 				currentMap.Value[key] = &entities.FactValueMap{Value: make(map[string]entities.FactValue)}
 			}
 
-			cursor = currentMap.Value[key].(*entities.FactValueMap)
+			cursor = currentMap.Value[key].(*entities.FactValueMap) //nolint:forcetypeassert
 		}
 	}
 
--- a/internal/factsengine/gatherers/corosynccmapctl.go
+++ b/internal/factsengine/gatherers/corosynccmapctl.go
@@ -86,14 +86,15 @@ func (s *CorosyncCmapctlGatherer) Gather(factsRequests []entities.FactRequest) (
 		return nil, CorosyncCmapCtlCommandError.Wrap(err.Error())
 	}
 
-	results := corosyncCmapctlOutputToMap(string(corosyncCmapctl))
+	corosyncCmapctlMap := corosyncCmapctlOutputToMap(string(corosyncCmapctl))
 
 	for _, factReq := range factsRequests {
 		var fact entities.Fact
 
-		value, err := results.GetValue(factReq.Argument)
-
-		if err == nil {
+		if len(factReq.Argument) == 0 {
+			log.Error(CorosyncCmapCtlMissingArgument.Message)
+			fact = entities.NewFactGatheredWithError(factReq, &CorosyncCmapCtlMissingArgument)
+		} else if value, err := corosyncCmapctlMap.GetValue(factReq.Argument); err == nil {
 			fact = entities.NewFactGatheredWithRequest(factReq, value)
 		} else {
 			gatheringError := CorosyncCmapCtlValueNotFound.Wrap(factReq.Argument)
--- a/internal/factsengine/gatherers/corosynccmapctl.go
+++ b/internal/factsengine/gatherers/corosynccmapctl.go
@@ -1,7 +1,7 @@
 package gatherers
 
 import (
-	"fmt"
+	"strings"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/agent/pkg/factsengine/entities"
@@ -45,6 +45,38 @@ func NewCorosyncCmapctlGatherer(executor utils.CommandExecutor) *CorosyncCmapctl
 	}
 }
 
+func corosyncCmapctlOutputToMap(corosyncCmapctlOutput string) *entities.FactValueMap {
+	outputMap := &entities.FactValueMap{Value: make(map[string]entities.FactValue)}
+	var cursor *entities.FactValueMap
+
+	for _, line := range strings.Split(corosyncCmapctlOutput, "\n") {
+		cursor = outputMap
+
+		value := strings.Split(line, "= ")[1]
+
+		pathAsString := strings.Split(line, " (")[0]
+		path := strings.Split(pathAsString, ".")
+
+		for i, key := range path {
+			currentMap := cursor
+
+			if i == len(path)-1 {
+				currentMap.Value[key] = entities.ParseStringToFactValue(value)
+
+				break
+			}
+
+			if _, found := currentMap.Value[key]; !found {
+				currentMap.Value[key] = &entities.FactValueMap{Value: make(map[string]entities.FactValue)}
+			}
+
+			cursor = currentMap.Value[key].(*entities.FactValueMap)
+		}
+	}
+
+	return outputMap
+}
+
 func (s *CorosyncCmapctlGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
 	facts := []entities.Fact{}
 	log.Infof("Starting %s facts gathering process", CorosyncCmapCtlGathererName)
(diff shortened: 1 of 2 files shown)


internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -45,6 +44,38 @@ func NewCorosyncCmapctlGatherer(executor utils.CommandExecutor) *CorosyncCmapctl
...
+				currentMap.Value[key] = entities.ParseStringToFactValue(value)
+
+				break
+			}
+
+			if _, found := currentMap.Value[key]; !found {
+				currentMap.Value[key] = &entities.FactValueMap{Value: make(map[string]entities.FactValue)}
+			}
+
+			cursor = currentMap.Value[key].(*entities.FactValueMap) //nolint:forcetypeassert
```
arbulu89, line 72: We shouldn't hide this assertion linting error. We should check if it's ok, and return an error if it cannot be asserted.

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -45,6 +44,38 @@ func NewCorosyncCmapctlGatherer(executor utils.CommandExecutor) *CorosyncCmapctl
...
+		for i, key := range path {
+			currentMap := cursor
+
+			if i == len(path)-1 {
+				currentMap.Value[key] = entities.ParseStringToFactValue(value)
+
+				break
+			}
+
+			if _, found := currentMap.Value[key]; !found {
```
arbulu89, line 68: I see a problem or at least a differentiation with other gatherers. Lists are not used.

When we have this:
```
nodelist.node.0.nodeid (u32) = 1
nodelist.node.0.ring0_addr (str) = 10.80.1.11
nodelist.node.1.nodeid (u32) = 2
nodelist.node.1.ring0_addr (str) = 10.80.1.12
```
We should treat the `0` and `1` values as list indexes instead of map keys.
Otherwise, when we want to loop them on rhai, we would need to loop them as maps, where the key is the index, which is strange.

I know that this makes the code a bit more difficult, but I think is the correct choice.
What do you think?

arbulu89 (changes requested):
Hey @jamie-suse ,
I think we need to improve the conversion to handle "lists" properly.
Let me know what do you think

internal/factsengine/gatherers/corosynccmapctl_test.go
```diff
@@ -136,55 +136,43 @@ func (suite *CorosyncCmapctlTestSuite) TestCorosyncCmapctlGatherer() {
 
 	factRequests := []entities.FactRequest{
 		{
-			Name:     "quorum_provider",
+			Name:     "nodes",
 			Gatherer: "corosync-cmapctl",
-			Argument: "quorum.provider",
+			Argument: "nodelist.node",
```
arbulu89, line 141: What happens if you add a new request with `nodelist.node.0` as argument?
I think it might fail as you store the "list" entries as maps

internal/factsengine/gatherers/corosynccmapctl.go
```diff
@@ -45,6 +44,38 @@ func NewCorosyncCmapctlGatherer(executor utils.CommandExecutor) *CorosyncCmapctl
...
+func corosyncCmapctlOutputToMap(corosyncCmapctlOutput string) *entities.FactValueMap {
+	outputMap := &entities.FactValueMap{Value: make(map[string]entities.FactValue)}
+	var cursor *entities.FactValueMap
+
+	for _, line := range strings.Split(corosyncCmapctlOutput, "\n") {
+		cursor = outputMap
+
+		value := strings.Split(line, "= ")[1]
+
+		pathAsString := strings.Split(line, " (")[0]
```
arbulu89, line 56: Other thing. I have seen that some values might include parenthesis, as they are string.
We might want to run this 2nd split using the `value` string instead of the `line`

internal/factsengine/gatherers/corosynccmapctl.go
rtorrero: We had a conversation with @fabriziosestito on Slack regarding this. I think you are right, a list with be ideal here but it increases the complexity of the code by quite a lot AFAICS. This way is just simpler, less code to maintain and just works.

internal/factsengine/gatherers/corosynccmapctl.go
fabriziosestito: The type assertion here could be ignored since we are 100% sure that the value is a `FactValueMap`
We could also do the error handling, as you proposed, but keep in mind that it's impossible to happen.

@CDimonaco what is the idiomatic way to do this?

internal/factsengine/gatherers/corosynccmapctl.go
fabriziosestito: Also, it will make it harder to document that when the config shows `0` and `1` you need to expect an array.
The output is presented as a nested map and this way what we return is always consistent.

internal/factsengine/gatherers/corosynccmapctl_test.go
fabriziosestito: Requesting `nodelist.node.0` returns a map containing two keys
`node_id` and `ring_addr` 

```golang
...
	Name: "nodes",
	Value: &entities.FactValueMap{
		Value: map[string]entities.FactValue{
			"nodeid":     &entities.FactValueInt{Value: 1},
			"ring0_addr": &entities.FactValueString{Value: "10.80.1.11"},
		},
	},
...
```

internal/factsengine/gatherers/corosynccmapctl.go
arbulu89: I'm not sure on this thinking...
I would expect a list honestly, as the are actually lists, where the number is the index.
Anyway, I already had talked with Jaime on this, and due some few inconsystencies in the output, we already agreed that using the map is the unique viable option.
This will force as to use `values()` in these entries in `rhai`, but well it is what it is

arbulu89: And mostly, the major inconsistency comes comparing in it with the other `corosync-conf` gatherer, which actually returns lists for the same values. So let's document this properly

internal/factsengine/gatherers/corosynccmapctl.go
arbulu89: Could you add a test on this?
The issue would come with this line: `runtime.totem.pg.mrp.srp.members.1.ip (str) = r(0) ip(10.80.1.11) `

internal/factsengine/gatherers/corosynccmapctl.go
arbulu89: I don't know...
We have done the assertion checking always in this cases, even knowing that it would get the same type 100% cases.
Anyway, I won't discuss on this

internal/factsengine/gatherers/corosynccmapctl.go
fabriziosestito: I totally see your point, as I am working on the documentation of the other gatherers and I realized that I was wrong, and the inconsistency will be in this gatherer. 
I am having second thoughts, so I am ok if you guys decide to go with the array solution.

>  due some few inconsistencies in the output we already agreed that using the map is the unique viable option

What are these inconsistencies about?

arbulu89: Outputs like this:
```
runtime.services.cmap.0.rx (u64) = 3
runtime.services.cmap.0.tx (u64) = 2
runtime.services.cmap.service_id (u16) = 0
```
Which at the end mix what it should be a list with other fields...

arbulu89: @jamie-suse by the way, could we add other test for this particular scenario?

rtorrero (commented):
Looks great @jamie-suse, well done!

Btw, what happens if in a FactRequest the argument finishes with a dot (.) after a valid key? 

e.g.: `nodelist.node.`

jamie-suse:
> Looks great @jamie-suse, well done!
> 
> Btw, what happens if in a FactRequest the argument finishes with a dot (.) after a valid key?
> 
> e.g.: `nodelist.node.`

Thanks @rtorrero, if a `FactRequest` is provided with a `Argument` that ends in a dot i.e. `"totem."`, then when the `Argument` is parsed, the trailing dot is removed and the value for `"totem"` is returned. Is there a problem with this behaviour?

fabriziosestito (approved):

nelsonkopliku (approved):
Great!

arbulu89 (approved):
Thanks @jamie-suse

PR #169, opened 2023-01-13
Add version comparison functionality for package_version

rtorrero:
This PR allows the `package_version` gatherer to be used to compare versions on the agent. This way, we can handle the complexity of comparing RPM package version strings to the OS using `zypper versioncmp` (see `zypper versioncmp --help` for more details).

Notice that this just adds to the existing functionality of requesting package version strings which should still work as it previously did. To request a version comparison, a package name together with a version string needs to be provided in the argument, separated by a coma  (`,`).

The returned value for such requests is a FactValueInt, containing:

 - A value of `0` if the provided version string matches the installed package version for the requested package
 - A value of `-1` if the provided version string is older that what's currently installed.
 - A value of `1` if the provided version string is newer than what's currently installed.

Example on how to collect such facts:

```
factRequests := []entities.FactRequest{
{
	Name:     "corosync_same_version",
	Gatherer: "package_version",
	Argument: "corosync,2.4.5",
	CheckID:  "check1",
},
{
	Name:     "corosync",
	Gatherer: "package_version",
	Argument: "corosync",
	CheckID:  "check2",
},
```

For such requests, the returned values should look like:
```
[]entities.Fact{
{
	Name:    "corosync_same_version",
	Value:   &entities.FactValueInt{Value: 0},
	CheckID: "check3",
},
{
	Name:    "corosync",
	Value:   &entities.FactValueString{Value: "2.4.5"},
	CheckID: "check1",
},
```
Notice how the first fact request asks for `corosync` against `2.4.5`, which will cause that an integer instead of a regular `FactValueString` with a version string. The second one, as it just asks for a package name will return a version string as it did before.

### DIFF
--- a/internal/factsengine/gatherers/packageversion.go
+++ b/internal/factsengine/gatherers/packageversion.go
@@ -16,9 +16,14 @@ const (
 
 // nolint:gochecknoglobals
 var (
-	PackageVersionCommandError = entities.FactGatheringError{
-		Type:    "package-version-cmd-error",
+	PackageVersionRpmCommandError = entities.FactGatheringError{
+		Type:    "package-version-rpm-cmd-error",
 		Message: "error while fetching package version",
+	}
+
+	PackageVersionZypperCommandError = entities.FactGatheringError{
+		Type:    "package-version-zypper-cmd-error",
+		Message: "error while executing zypper",
 	}
 
 	PackageVersionMissingArgument = entities.FactGatheringError{
@@ -96,14 +101,14 @@ func executeZypperVersionCmpCommand(
 ) (int, *entities.FactGatheringError) {
 	zypperOutput, err := executor.Exec("/usr/bin/zypper", "--terse", "versioncmp", comparedVersion, installedVersion)
 	if err != nil {
-		gatheringError := PackageVersionCommandError.Wrap(err.Error())
+		gatheringError := PackageVersionZypperCommandError.Wrap(err.Error())
 		log.Error(gatheringError)
 		return invalidVersionCompare, gatheringError
 	}
 
 	result, err := strconv.ParseInt(string(zypperOutput), 10, 32)
 	if err != nil {
-		gatheringError := PackageVersionCommandError.Wrap(err.Error())
+		gatheringError := PackageVersionRpmCommandError.Wrap(err.Error())
 		log.Error(gatheringError)
 		return invalidVersionCompare, gatheringError
 	}
@@ -117,7 +122,7 @@ func executeRpmVersionRetrieveCommand(
 ) (string, *entities.FactGatheringError) {
 	rpmOutput, err := executor.Exec("/usr/bin/rpm", "-q", "--qf", "%{VERSION}", packageName)
 	if err != nil {
-		gatheringError := PackageVersionCommandError.Wrap(string(rpmOutput) + err.Error())
+		gatheringError := PackageVersionRpmCommandError.Wrap(string(rpmOutput) + err.Error())
 		log.Error(gatheringError)
 		return "", gatheringError
 	}
--- a/internal/factsengine/gatherers/packageversion.go
+++ b/internal/factsengine/gatherers/packageversion.go
@@ -1,6 +1,9 @@
 package gatherers
 
 import (
+	"strconv"
+	"strings"
+
 	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/agent/pkg/factsengine/entities"
 	"github.com/trento-project/agent/pkg/utils"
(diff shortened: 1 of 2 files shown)


internal/factsengine/gatherers/packageversion.go
```diff
@@ -50,19 +54,85 @@ func (g *PackageVersionGatherer) Gather(factsRequests []entities.FactRequest) ([
...
+		if strings.Contains(factReq.Argument, ",") {
+			arguments := strings.SplitN(factReq.Argument, ",", 2)
+			packageName = arguments[0]
+			requestedVersion = arguments[1]
+		}
+
+		version, err := executeRpmVersionRetrieveCommand(g.executor, packageName)
 		if err != nil {
-			gatheringError := PackageVersionCommandError.Wrap(factReq.Argument)
+			gatheringError := PackageVersionCommandError.Wrap(packageName)
```
arbulu89, line 67: Do we need to wrap the error again? It is already wrapped and logged in `executeRpmVersionRetrieveCommand`.
It loos redundant

internal/factsengine/gatherers/packageversion.go
```diff
@@ -50,19 +54,85 @@ func (g *PackageVersionGatherer) Gather(factsRequests []entities.FactRequest) ([
...
 			log.Error(gatheringError)
 			fact = entities.NewFactGatheredWithError(factReq, gatheringError)
-		} else {
-			fact = entities.NewFactGatheredWithRequest(factReq, &entities.FactValueString{Value: (string(version))})
+			facts = append(facts, fact)
+			continue
 		}
 
+		if requestedVersion != "" {
+			installedVersion, err := executeRpmVersionRetrieveCommand(g.executor, packageName)
```
arbulu89, line 75: We don't need to run this again. You already have the version in `version`.
This way you can remove the next error checking as well

internal/factsengine/gatherers/packageversion.go
```diff
@@ -50,19 +54,85 @@ func (g *PackageVersionGatherer) Gather(factsRequests []entities.FactRequest) ([
...
+			if err != nil {
+				gatheringError := PackageVersionCommandError.Wrap(packageName)
+				log.Error(gatheringError)
+				fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+				facts = append(facts, fact)
+				continue
+			}
+			comparisonResult, err := executeZypperVersionCmpCommand(g.executor, installedVersion, requestedVersion)
+			if err != nil {
+				gatheringError := PackageVersionCommandError.Wrap(packageName)
```
arbulu89, line 85: The same here. I don't think you need to wrap this

internal/factsengine/gatherers/packageversion.go
```diff
@@ -50,19 +54,85 @@ func (g *PackageVersionGatherer) Gather(factsRequests []entities.FactRequest) ([
...
+				fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+				facts = append(facts, fact)
+				continue
+			}
+			fact = entities.NewFactGatheredWithRequest(factReq, &entities.FactValueInt{Value: comparisonResult})
+			facts = append(facts, fact)
+			continue
+		}
+
+		fact = entities.NewFactGatheredWithRequest(factReq, &entities.FactValueString{Value: (version)})
```
arbulu89, line 96: What are these parenthesis in the `version` for?

internal/factsengine/gatherers/verifypassword.go
```diff
@@ -73,7 +73,7 @@ func (g *VerifyPasswordGatherer) Gather(factsRequests []entities.FactRequest) ([
 
 		crypter := sha512crypt.New()
 		isPasswordWeak := false
-		var gatheringError *entities.FactGatheringError = nil
+		var gatheringError *entities.FactGatheringError
```
arbulu89, line 76: Not a big fun of adding this off topic changes in this PR...
I wouldn't include them

internal/factsengine/gatherers/packageversion_test.go
```diff
@@ -71,10 +71,16 @@ func (suite *PackageVersionTestSuite) TestPackageVersionGathererNoArgumentProvid
 }
 
 func (suite *PackageVersionTestSuite) TestPackageVersionGather() {
-	suite.mockExecutor.On("Exec", "rpm", "-q", "--qf", "%{VERSION}", "corosync").Return(
+	suite.mockExecutor.On("Exec", "/usr/bin/rpm", "-q", "--qf", "%{VERSION}", "corosync").Return(
```
arbulu89, line 74: Can we add some new test to check that errors during the comparison command work as expected? In fact, the test code doesn't check any scenario for failed commands

arbulu89 (changes requested):
I see some redundant call to get the version of the package.
PD: I wouldn't add not related changes to the PR

nelsonkopliku (approved):
Thanks!

arbulu89 (approved):
Thanks @rtorrero

PR #171, opened 2023-01-17
Handle `CorosyncCmapctlGatherer` receiving empty lines

jamie-suse:
Currently, `CorosyncCmapctlGatherer` is failing when it receives an empty line in the output from `corosync-cmapctl`. This is because it attempts to perform string splitting operations on the empty line.

This PR adds a check for empty lines and omits processing them. Also it adds an ending newline character to the test data file `corosynccmap-ctl.output` so that it will be like the output from `corosync-cmapctl`.


nelsonkopliku (approved):
Damn newlines :smile:

arbulu89 (approved):
That's why we used regex in the previous implementation hehe

PR #172, opened 2023-01-17
Unhide facts service url flag

arbulu89:
First step on the old legacy runner removal process.
Make the `facts-service-url` flag visible and the facts engine itself mandatory (removing the `factsengine` flag).
It includes all the boilerplate around the installation.

Next step is to remove the `ssh-address` flag which would be the last dependency with the old runner.


jamie-suse (approved):

dottorblaster (approved):
LGTM

PR #176, opened 2023-01-24
Pretty print fact values

dottorblaster:
This PR introduces pretty printing for fact values into the CLI, formatting maps and lists as Rhai object maps and arrays.

### DIFF
--- a/pkg/factsengine/entities/fact_value.go
+++ b/pkg/factsengine/entities/fact_value.go
@@ -1,10 +1,14 @@
 package entities
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"math"
 	"strconv"
 	"strings"
+
+	"github.com/pkg/errors"
 )
 
 // nolint:gochecknoglobals
@@ -191,4 +195,29 @@ func getValue(fact FactValue, values []string) (FactValue, error) {
 	default:
 		return value, nil
 	}
+}
+
+func Prettify(fact FactValue) (string, error) {
+	if fact == nil {
+		return "()", nil
+	}
+
+	interfaceValue := fact.AsInterface()
+
+	jsonResult, err := json.Marshal(interfaceValue)
+	if err != nil {
+		return "", errors.Wrap(err, "Error building the response")
+	}
+
+	var prettyfiedJSON bytes.Buffer
+	if err := json.Indent(&prettyfiedJSON, jsonResult, "", "  "); err != nil {
+		return "", errors.Wrap(err, "Error indenting the json data")
+	}
+
+	prettifiedJSONString := prettyfiedJSON.String()
+
+	rhaiValue := strings.ReplaceAll(prettifiedJSONString, "{", "#{")
+	rhaiValue = strings.ReplaceAll(rhaiValue, "null", "()")
+
+	return rhaiValue, nil
 }
--- a/pkg/factsengine/entities/facts_gathered.go
+++ b/pkg/factsengine/entities/facts_gathered.go
@@ -1,6 +1,10 @@
 package entities
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+)
 
 const (
 	FactsGathererdEventSource = "https://github.com/trento-project/agent"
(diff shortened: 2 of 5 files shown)


angelabriel:
:heart: Hey awesome, I really like this idea. Makes it so much easier to take the results from the gatherers to the playground to try out some 'wild' ideas :smiley:

dottorblaster:
@angelabriel original idea is by @fabriziosestito, I'm just an executor :D All the love goes to him :smile:

pkg/factsengine/entities/fact_value.go
```diff
@@ -98,6 +102,10 @@ func (v *FactValueString) AsInterface() interface{} {
 	return v.Value
 }
 
+func (v *FactValueString) PrettyPrint() (string, error) {
```
fabriziosestito, line 105: leftover?

pkg/factsengine/entities/facts_gathered.go
```diff
@@ -36,6 +40,16 @@ func (e *FactGatheringError) Wrap(msg string) *FactGatheringError {
 	}
 }
 
+func (e *Fact) PrettyPrint() (string, error) {
```
fabriziosestito, line 43: maybe we should cal this Prettify as well since it does not print

pkg/factsengine/entities/facts_gathered_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+type FactsGatheredTestSuite struct {
+	suite.Suite
+}
+
+func TestFactsGatheredTestSuite(t *testing.T) {
+	suite.Run(t, new(FactsGatheredTestSuite))
+}
+
+func (suite *FactsGatheredTestSuite) TestFactValuePrettify() {
```
fabriziosestito, line 18: this is testing FactsPrettyPrint not FactValuePrettify

fabriziosestito (changes requested):

pkg/factsengine/entities/facts_gathered.go
```diff
@@ -36,6 +40,16 @@ func (e *FactGatheringError) Wrap(msg string) *FactGatheringError {
 	}
 }
 
+func (e *Fact) Prettify() (string, error) {
+	prettifiedValue, err := Prettify(e.Value)
+	if err != nil {
+		return "", errors.Wrap(err, "Error pretty-printing fact value data")
```
arbulu89, line 46: Maybe better the error being `Error prettifying fact value data`

pkg/factsengine/entities/fact_value.go
```diff
@@ -192,3 +196,28 @@ func getValue(fact FactValue, values []string) (FactValue, error) {
 		return value, nil
 	}
 }
+
+func Prettify(fact FactValue) (string, error) {
```
arbulu89, line 200: Should this maybe be a private function?
And it could've been perfectly a struct function as well: 
`func (f *FactValue) Prettify() (string, error)`

arbulu89 (approved):
It looks great man!
Nice work

pkg/factsengine/entities/fact_value.go
dottorblaster: This was the same way I tried to pursue at the beginning, but it requires a method to be implemented for each implementation of `FactValue`. Instead, this way the function just works :tm: with every concrete type

pkg/factsengine/entities/facts_gathered.go
dottorblaster: Done :+1:

pkg/factsengine/entities/fact_value.go
fabriziosestito: @arbulu89 you cannot add methods to an interface

fabriziosestito (approved):

pkg/factsengine/entities/fact_value.go
CDimonaco: The function is better, you could achieve a semi functional result with a dedicated struct with the method implemented and then embedding with some hassle in every struct you want to prettify

But honestly I don't think it's very ideomatic, seems like some sort of oop abstract superclass with some default methods

CDimonaco (approved):

PR #193, opened 2023-02-28
Add ability to detect if running on `VMware` system

jamie-suse:


jamie-suse:
While making this PR, I found I was writing a lot of repeated code. Thoughts on a possible refactor?

internal/core/cloud/metadata.go
```diff
@@ -14,6 +14,7 @@ const (
 	GCP     = "gcp"
 	Nutanix = "nutanix"
 	KVM     = "kvm"
+	VMWARE  = "vmware"
```
arbulu89, line 17: Maybe bettter `VMware` as the constant name

arbulu89 (commented):
If we want to reduce the amount of times the `systemd-detect-virt` command is run, we could maybe create a new function like `systemdDetectVirt` which returns the command output value and then:

1. have some map like `{kvm: true, vmware: true}` or similar. Once the command is executed you could look up in the map, and if there is a value you know which is the detected one. In the other hand, you could return nil.

2. or simply continue having `identifyKVM` and `identifyVMware` functions, but using the command output as parameter and simply do the comparison there.

Take some time to think about other solutions as well.
Anyway, the current code should work, and even with high level of duplicated code, I find it easy to navigate

CDimonaco (commented):
Agree with @arbulu89 the code duplication and the command execution is really duplicated, the code could be a lot improved even in term of readability.

BUT, this is not the scope of this pr, the code works so we can think about later :D

rtorrero (approved):
Code looks good, green light from my side and we can later tackle how to reduce the duplication of code.

CDimonaco (approved):

PR #202, opened 2023-03-14
update spec file

stefanotorresi:
- refresh copyright notice
- remove golang-packaging build dependency, which is no longer needed
- make summary and descriptions more to the point


dottorblaster (approved):
LGTM

PR #204, opened 2023-03-14
Parse durations in cibadmin gatherer 

fabriziosestito:
Parses durations string in cibadmin gatherer


arbulu89 (approved):
Great thanks!

nelsonkopliku (approved):

arbulu89:
@fabriziosestito We should update now the gatherers docs in wand project.
This are the valid units `"ns", "us" (or "µs"), "ms", "s", "m", "h"`

PR #227, opened 2023-05-24
Send an empty payload if a cluster was not found

fabriziosestito:
This PR sends an empty payload if a cluster was not found, instead of returning from the discovery loop.
This is useful in the delta deregistration to remove a host from a cluster if previously it was part of it.

### DIFF
--- a/internal/core/cluster/cluster.go
+++ b/internal/core/cluster/cluster.go
@@ -61,7 +61,7 @@ func Md5sumFile(filePath string) (string, error) {
 	return hex.EncodeToString(hash.Sum(nil)), nil
 }
 
-func NewCluster() (Cluster, error) {
+func NewCluster() (*Cluster, error) {
 	return NewClusterWithDiscoveryTools(&DiscoveryTools{
 		CibAdmPath:      cibAdmPath,
 		CrmmonAdmPath:   crmmonAdmPath,
@@ -71,8 +71,16 @@ func NewCluster() (Cluster, error) {
 	})
 }
 
-func NewClusterWithDiscoveryTools(discoveryTools *DiscoveryTools) (Cluster, error) {
-	var cluster = Cluster{
+func NewClusterWithDiscoveryTools(discoveryTools *DiscoveryTools) (*Cluster, error) {
+	commandExecutor := utils.Executor{}
+	cibParser := cib.NewCibAdminParser(discoveryTools.CibAdmPath)
+
+	cibConfig, err := cibParser.Parse()
+	if err != nil {
+		return nil, err
+	}
+
+	var cluster = &Cluster{
 		Cib:      cib.Root{},    //nolint
 		Crmmon:   crmmon.Root{}, //nolint
 		SBD:      SBD{},         //nolint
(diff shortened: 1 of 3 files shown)


internal/discovery/cluster.go
```diff
@@ -38,9 +38,16 @@ func (c ClusterDiscovery) GetInterval() time.Duration {
 // Execute one iteration of a discovery and publish the results to the collector
 func (c ClusterDiscovery) Discover() (string, error) {
 	cluster, err := cluster.NewCluster()
+
 	if err != nil {
 		log.Debugf("Error creating the cluster data object: %s", err)
-		return "No HA cluster discovered on this host", nil
+	}
+
+	// If no cluster is found, the discovery payload is sent anyway.
```
arbulu89, line 46: I don't know if we should print this here.
In the next lines we would have a strange `"Cluster with name:  successfully discovered"` in this scenario.
What about moving this to the end of the file like we do for the SAPSystem discovery:
```
if cluster == nil {
    return "No HA cluster discovered on this host", nil
}

return fmt.Sprintf("Cluster with name: %s successfully discovered", cluster.Name), nil
```

arbulu89 (changes requested):
A little change request

internal/discovery/cluster.go
fabriziosestito: oh right i;ve missed this

arbulu89:
@fabriziosestito As talked offline, with the current change this is the payload that is generated:
```
{"agent_id":"xxx","discovery_type":"ha_cluster_discovery","payload":{"Cib":{"Configuration":{"CrmConfig":{"ClusterProperties":null},"Nodes":null,"Resources":{"Primitives":null,"Masters":null,"Clones":null,"Groups":null},"Constraints":{"RscLocations":null}}},"Crmmon":{"Version":"","Summary":{"Nodes":{"Number":0},"LastChange":{"Time":""},"Resources":{"Number":0,"Disabled":0,"Blocked":0},"ClusterOptions":{"StonithEnabled":false}},"Nodes":null,"NodeAttributes":{"Nodes":null},"NodeHistory":{"Nodes":null},"Resources":null,"Clones":null,"Groups":null},"SBD":{"Devices":null,"Config":null},"Id":"","Name":"cib-bootstrap-options-cluster-name","DC":false,"Provider":""}}
```
I think the `cluster` variable is not `nil` in this case.

internal/core/cluster/cluster.go
```diff
@@ -71,8 +71,16 @@ func NewCluster() (Cluster, error) {
 	})
 }
 
-func NewClusterWithDiscoveryTools(discoveryTools *DiscoveryTools) (Cluster, error) {
-	var cluster = Cluster{
+func NewClusterWithDiscoveryTools(discoveryTools *DiscoveryTools) (*Cluster, error) {
+	commandExecutor := utils.Executor{}
```
arbulu89, line 75: Why have you changed the order of the things here?
Was something wrong?
Or is it simply to avoid the declaration of a not needed var?

arbulu89 (approved):
Good!
Now we have:
```
{"agent_id":"xxx","discovery_type":"ha_cluster_discovery","payload":null}
```
If sending `null` in the payload is good enough for the backend (you might know more on that than me hehe), the code works as expected

internal/core/cluster/cluster.go
fabriziosestito: > Or is it simply to avoid the declaration of a not needed var?

this

PR #236, opened 2023-08-07
Remove not needed certificates

nelsonkopliku:
Just some clean up of not needed certificates coming from when we had mTLS.


arbulu89 (approved):
Thanks!

PR #243, opened 2023-08-22
Ignore extra output when comparing versions with zypper

nelsonkopliku:
This PR makes sure to ignore extra output of the zypper versioncmp command so that we can only take into account the actual comparinson result.

ie when the /etc/products.d/baseproduct symlink is missing a warning is returned in the output of a version comparison, breaking the parsing of the comparison value (expected to be -1, 0, 1)

```bash
$ zypper --terse versioncmp 1.5.2 1.5.2
Warning: The /etc/products.d/baseproduct symlink is dangling or missing!
The link must point to your core products .prod file in /etc/products.d.

0
```

### DIFF
--- a/internal/factsengine/gatherers/packageversion.go
+++ b/internal/factsengine/gatherers/packageversion.go
@@ -117,7 +117,10 @@ func executeZypperVersionCmpCommand(
 	}
 
 	versionCmpResult := strings.TrimRight(string(zypperOutput), "\n")
-	result, err := strconv.ParseInt(versionCmpResult, 10, 32)
+	outputParts := strings.Split(versionCmpResult, "\n")
+	comparisonResult := outputParts[len(outputParts)-1]
+
+	result, err := strconv.ParseInt(comparisonResult, 10, 32)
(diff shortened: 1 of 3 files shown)


internal/factsengine/gatherers/packageversion.go
```diff
@@ -117,7 +117,10 @@ func executeZypperVersionCmpCommand(
 	}
 
 	versionCmpResult := strings.TrimRight(string(zypperOutput), "\n")
-	result, err := strconv.ParseInt(versionCmpResult, 10, 32)
+	outputParts := strings.Split(versionCmpResult, "\n")
+	comparisonResult := outputParts[len(outputParts)-1]
```
nelsonkopliku, line 121: Just a doubt here: should we guard against `outputParts` being an empty slice? I guess accessing the last element of an empty slice might break something lol. However not sure that could even happen.

dottorblaster (approved):
I couldn't help asking myself "what if we get an empty string?"

but still, LGTM

PR #253, opened 2023-09-13
Add saptune discovery

rtorrero:
Add initial implementation of the `saptune --format json status` and expose this through the collector

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -47,6 +47,7 @@ func NewAgent(config *Config) (*Agent, error) {
 		discovery.NewCloudDiscovery(collectorClient, *config.DiscoveriesConfig),
 		discovery.NewSubscriptionDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
 		discovery.NewHostDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
+		discovery.NewSaptuneDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
 	}
 
 	agent := &Agent{
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -47,7 +47,7 @@ func NewAgent(config *Config) (*Agent, error) {
 		discovery.NewCloudDiscovery(collectorClient, *config.DiscoveriesConfig),
 		discovery.NewSubscriptionDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
 		discovery.NewHostDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
-		discovery.NewSaptuneDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
+		discovery.NewSaptuneDiscovery(collectorClient, *config.DiscoveriesConfig),
 	}
 
 	agent := &Agent{
--- a/internal/core/saptune/saptune.go
+++ b/internal/core/saptune/saptune.go
@@ -73,13 +73,5 @@ func (s *Saptune) RunCommandJSON(args ...string) ([]byte, error) {
 	}
 
 	prependedArgs := append([]string{"--format", "json"}, args...)
-	log.Infof("Running saptune command: saptune %v", args)
-	output, err := s.executor.Exec("saptune", prependedArgs...)
-	if err != nil {
-		log.Debugf(err.Error())
-	}
-	log.Debugf("saptune output: %s", string(output))
-	log.Infof("Saptune command executed")
-
-	return output, nil
+	return s.RunCommand(prependedArgs...)
 }
--- a/internal/core/saptune/saptune.go
+++ b/internal/core/saptune/saptune.go
@@ -1,8 +1,6 @@
 package saptune
 
 import (
-	"encoding/json"
-
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"github.com/trento-project/agent/pkg/utils"
@@ -19,53 +17,8 @@ const (
 )
 
 type Saptune struct {
-	Version		*string				`json:"version"`
-	Commands	[]SaptuneOutput		`json:"commands"`
-}
-type SaptuneOutput struct {
-	Schema      string    `json:"$schema"`
-	PublishTime string    `json:"publish time"`
-	Argv        string    `json:"argv"`
-	Pid         int       `json:"pid"`
-	Command     string    `json:"command"`
-	ExitCode    int       `json:"exit code"`
-	Result      Result    `json:"result"`
-	Messages    []Message `json:"messages"`
-}
-
-type Result struct {
-	Services                 Services `json:"services"`
-	SystemdSystemState       string   `json:"systemd system state"`
-	TuningState              string   `json:"tuning state"`
-	Virtualization           string   `json:"virtualization"`
-	ConfiguredVersion        string   `json:"configured version"`
-	PackageVersion           string   `json:"package version"`
-	SolutionEnabled          []string `json:"Solution enabled"`
-	NotesEnabledBySolution   []string `json:"Notes enabled by Solution"`
-	SolutionApplied          []string `json:"Solution applied"`
-	NotesAppliedBySolution   []string `json:"Notes applied by Solution"`
-	NotesEnabledAdditionally []string `json:"Notes enabled additionally"`
-	NotesEnabled             []string `json:"Notes enabled"`
-	NotesApplied             []string `json:"Notes applied"`
-	Staging                  Staging  `json:"staging"`
-	RememberMessage          string   `json:"remember message"`
-}
-
-type Services struct {
-	Saptune []string `json:"saptune"`
-	Sapconf []string `json:"sapconf"`
-	Tuned   []string `json:"tuned"`
-}
-
-type Staging struct {
-	StagingEnabled  bool     `json:"staging enabled"`
-	NotesStaged     []string `json:"Notes staged"`
-	SolutionsStaged []string `json:"Solutions staged"`
-}
-
-type Message struct {
-	Priority string `json:"priority"`
-	Message  string `json:"message"`
+	Version  string `json:"version"`
+	executor utils.CommandExecutor
 }
 
 func getSaptuneVersion(commandExecutor utils.CommandExecutor) (string, error) {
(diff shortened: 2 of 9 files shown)


internal/agent/agent.go
```diff
@@ -47,6 +47,7 @@ func NewAgent(config *Config) (*Agent, error) {
 		discovery.NewCloudDiscovery(collectorClient, *config.DiscoveriesConfig),
 		discovery.NewSubscriptionDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
 		discovery.NewHostDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
+		discovery.NewSaptuneDiscovery(collectorClient, config.InstanceName, *config.DiscoveriesConfig),
```
arbulu89, line 50: Is this `InstanceName` used for something? I didn't see any usage in the discovery struct

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+
+const (
+	MinimalSaptuneVersion = "v3.1.0"
+)
+
+type Saptune struct {
+	Version  *string         `json:"version"`
+	Commands []SaptuneOutput `json:"commands"`
+}
+type SaptuneOutput struct {
```
arbulu89: Do we need all this structs here?
I mean, we are just bypassing the json, we don't use them at all internally

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+var (
+	ErrSaptuneVersionUnknown = errors.New("could not determine saptune version")
+	ErrUnsupportedSaptuneVer = errors.New("saptune version is not supported")
+)
+
+const (
+	MinimalSaptuneVersion = "v3.1.0"
+)
+
+type Saptune struct {
```
arbulu89, line 21: Most probably you want to revisit this to use the new format with `package_version`, `saptune_installed`, etc

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+type Message struct {
+	Priority string `json:"priority"`
+	Message  string `json:"message"`
+}
+
+func getSaptuneVersion(commandExecutor utils.CommandExecutor) (string, error) {
+	log.Info("Requesting Saptune version...")
+	versionOutput, err := commandExecutor.Exec("rpm", "-q", "--qf", "%{VERSION}", "saptune")
+	if err != nil {
+		return "", ErrSaptuneVersionUnknown
```
arbulu89, line 75: Let's wrap the coming `err` here, otherwise we just get the generic error, without any context of why it failed

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+	SolutionsStaged []string `json:"Solutions staged"`
+}
+
+type Message struct {
+	Priority string `json:"priority"`
+	Message  string `json:"message"`
+}
+
+func getSaptuneVersion(commandExecutor utils.CommandExecutor) (string, error) {
+	log.Info("Requesting Saptune version...")
```
arbulu89, line 72: Strange to print this here if we are not going to print as `Info` the version afterwards

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+		return "", ErrSaptuneVersionUnknown
+	}
+
+	log.Debugf("saptune version output: %s", string(versionOutput))
+
+	return string(versionOutput), nil
+}
+
+func isSaptuneVersionSupported(version string) bool {
+
```
arbulu89, line 84: Let's remove this initial empty line

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+		return Saptune{Version: nil, Commands: []Output{}}, err
+	}
+
+	if !isSaptuneVersionSupported(saptuneVersion) {
+		return Saptune{Version: &saptuneVersion, Commands: []Output{}}, ErrUnsupportedSaptuneVer
+	}
+
+	log.Info("Requesting Saptune status...")
+	output, err := commandExecutor.Exec("saptune", "--format", "json", "status")
+	if err != nil {
```
arbulu89, line 102: Be careful with this. Saptune has some predefined exit codes, from 0 to 4 i think, where they still return the correct format, but it is determined if saptune tuning is working or not

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,120 @@
...
+		return Saptune{Version: &saptuneVersion, Commands: []Output{}},
+			errors.Wrap(err, "unexpected error while calling saptune")
+	}
+
+	log.Debugf("saptune output: %s", string(output))
+
+	var saptuneOutput Output
```
arbulu89, line 109: Do we need to unmarshall, why don't we send directly the json output in the discovery payload?

arbulu89 (commented):
Hey @rtorrero ,
Some initial comments.
Besides that, I wonder how this code could be reusable in a gatherer. I would maybe expect some `RunSaptuneCommand(cmd, format)`, where you pass the command you want to execute and maybe the format.

internal/core/saptune/saptune.go
rtorrero: No need and can be removed. I thought they might be used in the future if we need to do something with the JSON output before sending it, but I can remove & re-add when needed.

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,71 @@
...
+
+	return Saptune{Version: saptuneVersion, executor: commandExecutor}, nil
+}
+
+func (s *Saptune) RunCommand(args ...string) ([]byte, error) {
+	if !isSaptuneVersionSupported(s.Version) {
+		return nil, ErrUnsupportedSaptuneVer
+	}
+
+	log.Info("Requesting Saptune status...")
```
arbulu89, line 56: These initial log messages are outdated, right?

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,71 @@
...
+
+	log.Infof("Saptune status discovered")
+	log.Infof("Running saptune command: saptune %v", args)
+	output, err := s.executor.Exec("saptune", args...)
+	log.Debugf("saptune output: %s", string(output))
+	if err != nil {
```
arbulu89, line 62: I don't think we should do any error handling here.
We should return output and err, directly.
As commented, in `saptune status`, having an error code doesn't mean that the command failed.
You can debug the output before returning

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,71 @@
...
+	saptuneVersion, err := getSaptuneVersion(commandExecutor)
+	if err != nil {
+		return Saptune{Version: ""}, err
+	}
+
+	return Saptune{Version: saptuneVersion, executor: commandExecutor}, nil
+}
+
+func (s *Saptune) RunCommand(args ...string) ([]byte, error) {
+	if !isSaptuneVersionSupported(s.Version) {
```
arbulu89, line 52: Are we sure we want to run this for every command?
If we had the `isJsonSupported` field, we could avoid running this. 
Actually, the code owner (us) could even avoid using `RunCommand`, as we know that json is not supported before hand

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,71 @@
...
+	ErrSaptuneVersionUnknown = errors.New("could not determine saptune version")
+	ErrUnsupportedSaptuneVer = errors.New("saptune version is not supported")
+)
+
+const (
+	MinimalSaptuneVersion = "v3.1.0"
+)
+
+type Saptune struct {
+	Version  string `json:"version"`
```
arbulu89, line 20: I don't think we need the json annotation.
This struct is internal, not meant to be marshalled

internal/core/saptune/saptune_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+type SaptuneTestSuite struct {
+	suite.Suite
+}
+
+func TestSaptuneTestSuite(t *testing.T) {
+	suite.Run(t, new(SaptuneTestSuite))
+}
+
+func (suite *SaptuneTestSuite) TestNewSaptune() {
```
arbulu89, line 19: I'm a bit confused with the tests.
I would like to have some tests for the `NewSaptune` and other tests for the `RunCommand`.
They don't need to be combined. For the `RunCommand` you can create the saptune struct of your convinience

internal/discovery/saptune.go
```diff
@@ -0,0 +1,64 @@
...
+	return d.id
+}
+
+func (d SaptuneDiscovery) GetInterval() time.Duration {
+	return d.interval
+}
+
+func (d SaptuneDiscovery) Discover() (string, error) {
+	saptuneRetriever, err := saptune.NewSaptune(utils.Executor{})
+	if err != nil {
```
arbulu89, line 45: We cannot do that. Even if we get some error here, we need to send a payload, with the `PackageVersion: null/empty string`, `SaptuneInstalled: false` and `Status: nil`

internal/discovery/saptune.go
```diff
@@ -0,0 +1,64 @@
...
+		return "", err
+	}
+
+	saptuneData, _ := saptuneRetriever.RunCommand("--format", "json", "status")
+
+	var saptunePayload = SaptuneDiscoveryPayload{
+		PackageVersion:   saptuneRetriever.Version,
+		SaptuneInstalled: saptuneRetriever.Version != "",
```
arbulu89, line 53: At this point `SaptuneInstalled` should be always `true`, right?
I shouldn't need a comparison

internal/discovery/saptune.go
```diff
@@ -0,0 +1,64 @@
...
+	return d.interval
+}
+
+func (d SaptuneDiscovery) Discover() (string, error) {
+	saptuneRetriever, err := saptune.NewSaptune(utils.Executor{})
+	if err != nil {
+		return "", err
+	}
+
+	saptuneData, _ := saptuneRetriever.RunCommand("--format", "json", "status")
```
arbulu89, line 49: We cannot ignore the error.
What if json format is not supported?
That's why I like to have the `IsJsonSupported` field.
Because you can check it here, and if it is false, send a payload with `status: nil` without even using `RunCommand`

arbulu89 (changes requested):
Hey @rtorrero 
Some feedback.
Consider adding a `IsJsonSupported` field. I think it will come really useful in the gatherers as well, as we can skip running command functions since the beginning

internal/discovery/saptune.go
arbulu89: Alright, I understood now.
You are redirecting the nil coming from the `RunCommand`, a bit ugly for me, but it should work

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,66 @@
...
+func isSaptuneVersionSupported(version string) bool {
+	compareOutput := semver.Compare(MinimalSaptuneVersion, "v"+version)
+
+	return compareOutput != 1
+}
+
+func NewSaptune(commandExecutor utils.CommandExecutor) (Saptune, error) {
+	saptuneVersion, err := getSaptuneVersion(commandExecutor)
+	if err != nil {
+		return Saptune{Version: ""}, err
```
arbulu89, line 46: I think you could simply return `Saptune{}`

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	return Saptune{Version: saptuneVersion, executor: commandExecutor, IsJSONSupported: isSaptuneVersionSupported(saptuneVersion)}, nil
```
arbulu89, line 49: Maybe better to store the result in a variable, just to improve readability

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (s *Saptune) RunCommand(args ...string) ([]byte, error) {
+	if !s.IsJSONSupported {
```
arbulu89, line 53: If we forbid non json options, we should hardcode the `--format json` in the command, so the user doesn't need to put it.

arbulu89 (commented):
Looking good!
Almost ready to merge, once we fix the linting things mostly

internal/core/saptune/saptune_test.go
```diff
@@ -0,0 +1,112 @@
...
+	mockCommand := new(mocks.CommandExecutor)
+
+	mockCommand.On("Exec", "rpm", "-q", "--qf", "%{VERSION}", "saptune").Return(
+		[]byte("3.1.0"), nil,
+	)
+
+	saptuneRetriever, err := NewSaptune(mockCommand)
+
+	expectedDetails := Saptune{
```
arbulu89, line 28: Maybe we could name this `expectedSaptune`. Wdyt?

internal/core/saptune/saptune.go
```diff
@@ -0,0 +1,85 @@
...
+	return output, nil
+}
+
+func (s *Saptune) RunCommandJSON(args ...string) ([]byte, error) {
+	if !s.IsJSONSupported {
+		return nil, ErrUnsupportedSaptuneVer
+	}
+
+	prependedArgs := append([]string{"--format", "json"}, args...)
+	log.Infof("Running saptune command: saptune %v", args)
```
arbulu89, line 76: We could reuse the main `RunCommand` once we have the new args.
Not a big deal

internal/discovery/saptune.go
```diff
@@ -0,0 +1,80 @@
...
+	case !saptuneRetriever.IsJSONSupported:
+		saptunePayload = SaptuneDiscoveryPayload{
+			PackageVersion:   saptuneRetriever.Version,
+			SaptuneInstalled: true,
+			Status:           nil,
+		}
+	default:
+		saptuneData, _ := saptuneRetriever.RunCommandJSON("status")
+		unmarshalled := make(map[string]interface{})
+		_ = json.Unmarshal(saptuneData, &unmarshalled)
```
arbulu89, line 64: I think we should capture the error and exit if there is any.
Just for safetiness. 
The status should always return  a json, but just in case...

arbulu89 (approved):
@rtorrero green light from my side.
It would be nice to run some manual integration test between agent and web to see that it works.
(remember that there is a pending PR in web which fixes the policy)

internal/discovery/saptune.go
rtorrero: The thing is that as you said, some `saptune` commands will return non-exit codes which still have valid data. If we exit we wouldn't return that JSON, right? or do you mean to simply:

```
		if err != nil {
			log.Debugf("Error while unmarshalling saptune status: %s", err)
		}
```
and thats it?

PR #255, opened 2023-09-19
Use json.RawMessage type to avoid unnecessary json.Unmarshal

rtorrero:
This PR changes the SaptuneDiscoveryPayload struct to avoid the usage of `interface{}` type and `json.Unmarshal`. Instead we use a `json.RawMessage` that prevents the payload from being quoted.

### DIFF
--- a/internal/discovery/saptune.go
+++ b/internal/discovery/saptune.go
@@ -20,9 +20,9 @@ type SaptuneDiscovery struct {
 }
 
 type SaptuneDiscoveryPayload struct {
-	PackageVersion   string      `json:"package_version"`
-	SaptuneInstalled bool        `json:"saptune_installed"`
-	Status           interface{} `json:"status"`
+	PackageVersion   string          `json:"package_version"`
+	SaptuneInstalled bool            `json:"saptune_installed"`
+	Status           json.RawMessage `json:"status"`
 }
 
 func NewSaptuneDiscovery(collectorClient collector.Client, config DiscoveriesConfig) Discovery {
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):

arbulu89 (commented):
Hey @rtorrero ,
What if the output of the command for some error is not json?
The `Unmarshal` would catch that

rtorrero:
> Hey @rtorrero , What if the output of the command for some error is not json? The `Unmarshal` would catch that

It shouldn't matter, since we aren't doing anything with the JSON itself (It should fail once we try to unmarshal or decode it). If I understood it correctly, for us `json.RawMessage` is just acting as an alias of `[]byte`.

arbulu89:
> > Hey @rtorrero , What if the output of the command for some error is not json? The `Unmarshal` would catch that
> 
> It shouldn't matter, since we aren't doing anything with the JSON itself (It should fail once we try to unmarshal or decode it). If I understood it correctly, for us `json.RawMessage` is just acting as an alias of `[]byte`.

Well, you check at least that it is a json format.
What happens if for some reason we get a non json output from the command?
Would we send that plain text in the `payload` field?
Because the collector will marshall this as json.
I just want to be sure that we don't panic in that scenario

rtorrero:
> > > Hey @rtorrero , What if the output of the command for some error is not json? The `Unmarshal` would catch that
> > 
> > 
> > It shouldn't matter, since we aren't doing anything with the JSON itself (It should fail once we try to unmarshal or decode it). If I understood it correctly, for us `json.RawMessage` is just acting as an alias of `[]byte`.
> 
> Well, you check at least that it is a json format. What happens if for some reason we get a non json output from the command? Would we send that plain text in the `payload` field? Because the collector will marshall this as json. I just want to be sure that we don't panic in that scenario

This would be caught by 

```
func (c *Collector) Publish(discoveryType string, payload interface{}) error {
	log.Debugf("Sending %s to data collector", discoveryType)

	requestBody, err := json.Marshal(map[string]interface{}{
		"agent_id":       c.config.AgentID,
		"discovery_type": discoveryType,
		"payload":        payload,
	})
	if err != nil {
		return err
	}
	...
```

in our collector.

We could catch it here if you prefer, but IMO that defeats a bit the purpose of this PR

CDimonaco (approved):

PR #256, opened 2023-09-20
Saptune gatherer

rtorrero:
This PR adds a new Saptune gatherer using the same saptune core code as the https://github.com/trento-project/agent/pull/253 (`saptune_discovery`). The whitelisted saptune arguments are the following:

```
var whitelistedArguments = map[string][]string{
	"status":          {"status", "--non-compliance-check"},
	"solution-verify": {"solution", "verify"},
	"solution-list":   {"solution", "list"},
	"note-verify":     {"note", "verify"},
	"note-list":       {"note", "list"},
}
```


internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+	return NewSaptuneGatherer(utils.Executor{})
+}
+
+func NewSaptuneGatherer(executor utils.CommandExecutor) *SaptuneGatherer {
+	return &SaptuneGatherer{
+		executor: executor,
+	}
+}
+
+func parseJSONToFactValue(jsonStr string) (entities.FactValue, error) {
```
arbulu89, line 64: Why `string` as input? 
We should be using `json.RawMessage` I guess

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+	// Unmarshal the JSON into an interface{} type.
+	var jsonData interface{}
+	if err := json.Unmarshal([]byte(jsonStr), &jsonData); err != nil {
+		return nil, err
+	}
+
+	// Convert the parsed jsonData into a FactValue using NewFactValue.
+	return entities.NewFactValueMod(jsonData)
```
arbulu89, line 72: I don't think we need any new function like this.
We already have other gatherers converting a map to `FactValueMap`.
Check the `xml.go` file for example

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (s *SaptuneGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting %s facts gathering process", SaptuneGathererName)
+	saptuneRetriever, _ := saptune.NewSaptune(s.executor)
+	for _, factReq := range factsRequests {
+			var fact entities.Fact
+		if len(factReq.Argument) == 0 {
```
arbulu89, line 81: I think a `switch/case` would be easier to read

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+			log.Error(SaptuneMissingArgument.Message)
+			fact = entities.NewFactGatheredWithError(factReq, &SaptuneMissingArgument)
```
arbulu89, line 83: We need to add a clause the return the not supported if `IsJSONSupported` if false

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+) (entities.FactValue, *entities.FactGatheringError) { 
+	internalArguments, ok := whitelistedArguments[argument]
+
+	if !ok {
+		gatheringError := SaptuneUnknownArgument.Wrap(internalArguments)
+		log.Error(gatheringError)
+		return nil, gatheringError
+	}
+
+	argList := strings.Split(internalArguments, " ")
```
arbulu89, line 108: Why this split? you already have the command you want to run in the map you created for the whitelist

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+	return facts, nil
+}
+
+func handleArgument(
+	saptuneRetriever *saptune.Saptune,
+	argument string,
+) (entities.FactValue, *entities.FactGatheringError) { 
+	internalArguments, ok := whitelistedArguments[argument]
+
+	if !ok {
```
arbulu89, line 102: I think it would be better to include this whitelisting logic together with the `len(arguments) == 0` logic.
At the end, we are validating the argument.
Some new `func validateArgument` kind of function would be nice

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,128 @@
...
+	if commandError != nil {
+		gatheringError := SaptuneCommandError.Wrap(commandError.Error())
+		log.Error(gatheringError)
+		return nil, gatheringError
+	}
+
+	return gatherFactsFromOutput(saptuneOutput)
+}
+
+func gatherFactsFromOutput(commandOutput []byte) (entities.FactValue, *entities.FactGatheringError) {
```
arbulu89, line 119: Do we need a new function for a `if/else` snippet?

arbulu89 (commented):
@rtorrero Some few things.
Besides that, i don't still see any "caching" system.
I mean, we must only run the commands for each argument once, so you need to check if they were executed, and store the output

internal/factsengine/gatherers/saptune.go
rtorrero: This is to avoid the default conversion that always converts "3" to a FactValueInt{Value: 3}. If we are ok with any quoted number to be converted to a FactValueInt, then we can remove it.

internal/factsengine/gatherers/saptune.go
rtorrero: Because when the arguments are passed to the executor they need to be split as an array. Calling `s.executor.Exec("saptune", "some space separated argument)` wouldn't work, right? it should be `("saptune", "some", "space", "separated", "argument")`

internal/factsengine/gatherers/saptune.go
arbulu89: We should be fine in my opinion, but if you want you could ask the checks team, as they will be the final users afterall

internal/factsengine/gatherers/saptune.go
arbulu89: You could try to store the command values as string in the whitelisted variable

internal/factsengine/gatherers/saptune.go
rtorrero: sure, can be removed

test/fixtures/gatherers/saptune-note-list.output
```diff
@@ -0,0 +1,2 @@
+
```
arbulu89, line 1: Just for purity we could remove this first empty line

rtorrero: I added it because I noticed the output of `saptune` adds it, but it shouldn't make any difference

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,140 @@
...
+	saptuneRetriever, err := saptune.NewSaptune(s.executor)
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+
+		internalArguments, ok := whitelistedArguments[factReq.Argument]
+		cachedFact, cacheHit := cachedFacts[factReq.Argument]
+
+		switch {
+		case err != nil:
```
arbulu89, line 81: We don't need this `SaptuneNotInstalled` and `SaptuneVersionUnsupported` in the main loop.
If the `Gather` function return them as error, they will be added to all requests.

```
facts := []entities.Fact{}
log.Infof("Starting %s facts gathering process", SaptuneGathererName)
saptuneRetriever, err := saptune.NewSaptune(s.executor)
if err != nil {
    return facts, &SaptuneNotInstalled
}

if !saptuneRetriever.IsJSONSupported {
    return facts, &SaptuneVersionUnsupported
}

...
```

internal/factsengine/gatherers/saptune.go
```diff
@@ -0,0 +1,140 @@
...
+	return facts, nil
+}
+
+func runCommand(saptuneRetriever *saptune.Saptune, arguments []string) (entities.FactValue, error) {
+	saptuneOutput, commandError := saptuneRetriever.RunCommandJSON(arguments...)
+	if commandError != nil {
+		return nil, commandError
+	}
+
+	log.Error(string(saptuneOutput))
```
arbulu89, line 130: I think this was a temporary log output. Let's remove it.
The same for the one in line `137`

arbulu89 (changes requested):
Good!
Once we have this minor things fixed we are ready to merge.
You can start working in the documentation of the gatherer in wanda if you want

internal/factsengine/gatherers/saptune.go
```diff
@@ -70,22 +70,21 @@ func (s *SaptuneGatherer) Gather(factsRequests []entities.FactRequest) ([]entiti
 	facts := []entities.Fact{}
 	log.Infof("Starting %s facts gathering process", SaptuneGathererName)
 	saptuneRetriever, err := saptune.NewSaptune(s.executor)
-
+	if err != nil {
+		return facts, &SaptuneNotInstalled
```
arbulu89, line 74: We could return `nil, &SaptuneNotInstalled` as well (the same for the next)
And even `SaptuneNotInstalled.Wrap(err.Error())` in this case, as we have the rpm command error

arbulu89 (approved):
Thank you @rtorrero 
Ready to merge from my side.
PD: added last nitpick comment if you want to have a look :see_no_evil:

PR #260, opened 2023-09-26
Add configuration options to NewFactValue

arbulu89:
Create `NewFactValueWithConf` to enable configuration of the `NewFactValue` function.
Initially only `SnakeCaseKeys` and `StringConversion` options are added

### DIFF
--- a/pkg/factsengine/entities/fact_value.go
+++ b/pkg/factsengine/entities/fact_value.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/iancoleman/strcase"
 	"github.com/pkg/errors"
 )
 
@@ -19,6 +20,18 @@ var ValueNotFoundError = FactGatheringError{
 	Message: "error getting value",
 }
 
+type Conf struct {
+	StringConversion bool // Enable string automatic conversion to to a numeric fact value
+	SnakeCaseKeys    bool // snake_case keys if the provided value is a map
+}
+
+func NewDefaultConf() Conf {
+	return Conf{
+		StringConversion: true,
+		SnakeCaseKeys:    false,
+	}
+}
+
 // FactValue represents a dynamically typed value which can be either
 // an int, a float, a string, a boolean, a recursive map[string] value, or a
 // list of values.
@@ -28,13 +41,13 @@ type FactValue interface {
 	AsInterface() interface{}
 }
 
-// NewFactValue constructs a FactValue from a nested interface.
-func NewFactValue(factInterface interface{}) (FactValue, error) {
+// NewFactValueWithConf constructs a FactValue from a nested interface with a provided configuration
+func NewFactValueWithConf(factInterface interface{}, conf *Conf) (FactValue, error) {
 	switch value := factInterface.(type) {
 	case []interface{}:
 		newList := []FactValue{}
 		for _, value := range value {
-			newValue, err := NewFactValue(value)
+			newValue, err := NewFactValueWithConf(value, conf)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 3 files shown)


pkg/factsengine/entities/fact_value.go
```diff
@@ -44,20 +57,34 @@ func NewFactValue(factInterface interface{}) (FactValue, error) {
...
 				return nil, err
 			}
+			if conf.SnakeCaseKeys {
+				key = strcase.ToSnake(key)
+			}
 			newMap[key] = newValue
 		}
 		return &FactValueMap{Value: newMap}, nil
-	case bool, int, int32, int64, uint, uint32, uint64, float32, float64, string:
+	case bool, int, int32, int64, uint, uint32, uint64, float32, float64:
```
arbulu89, line 70: PD: I tried to split the case entries for the current `int` and `float` adding more `case` clauses, and avoid using the `ParseStringToFactValue` function for them, but it slightly changes the behaviour. For example, now, float numbers without decimals are converted to `int`, but if we get `float64` in the case, we would convert it to `float`.

This is how it would look like:
```
	case bool:
		return &FactValueBool{Value: value}, nil
	case int:
		return &FactValueInt{Value: value}, nil
	case int32:
		return &FactValueInt{Value: int(value)}, nil
	case int64:
		return &FactValueInt{Value: int(value)}, nil
	case uint:
		return &FactValueInt{Value: int(value)}, nil
	case uint32:
		return &FactValueInt{Value: int(value)}, nil
	case uint64:
		return &FactValueInt{Value: int(value)}, nil
	case float32:
		return &FactValueFloat{Value: float64(value)}, nil
	case float64:
		return &FactValueFloat{Value: value}, nil
	case string:
		if conf.StringConversion {
			return ParseStringToFactValue(value), nil
		}
		return &FactValueString{Value: value}, nil
```
But we might have some issues converting `uint64` to `int` in some corner case.
Wdyt?

rtorrero: Does this behaviour change affect any of the existing gatherers?

arbulu89: Yes, values like `30s` (which we have in cibadmin gatherer) pass to be of float type instead of int type

pkg/factsengine/entities/fact_value.go
```diff
@@ -44,20 +57,34 @@ func NewFactValue(factInterface interface{}) (FactValue, error) {
 	case map[string]interface{}:
 		newMap := make(map[string]FactValue)
 		for key, mapValue := range value {
-			newValue, err := NewFactValue(mapValue)
+			newValue, err := NewFactValueWithConf(mapValue, conf)
 			if err != nil {
 				return nil, err
 			}
+			if conf.SnakeCaseKeys {
+				key = strcase.ToSnake(key)
```
CDimonaco, line 65: Can we use a new scoped variable, not assigning to the key moving pointer in the for loop?

pkg/factsengine/entities/fact_value.go
```diff
@@ -44,20 +57,34 @@ func NewFactValue(factInterface interface{}) (FactValue, error) {
...
+			return ParseStringToFactValue(value), nil
+		}
+		return &FactValueString{Value: value}, nil
 	default:
 		return nil, fmt.Errorf("invalid type: %T for value: %v", value, factInterface)
 	}
 }
 
+// NewFactValue constructs a FactValue from a nested interface.
+func NewFactValue(factInterface interface{}) (FactValue, error) {
```
CDimonaco, line 83: Could we have this constructor named `NewFactValueWithDefaultConf`? In this way we can explicitely say that we are using this constructor as a proxy with the new factValue with the default configuration.

We don't have real optional values in golang and having this naming conventions helps us to recognize this proxies.

pkg/factsengine/entities/fact_value.go
```diff
@@ -28,13 +41,13 @@ type FactValue interface {
 	AsInterface() interface{}
 }
 
-// NewFactValue constructs a FactValue from a nested interface.
-func NewFactValue(factInterface interface{}) (FactValue, error) {
+// NewFactValueWithConf constructs a FactValue from a nested interface with a provided configuration
+func NewFactValueWithConf(factInterface interface{}, conf *Conf) (FactValue, error) {
```
CDimonaco, line 45: After the renaming of the proxy constructor we can safely revert that function to `NewFactValue`

CDimonaco (changes requested):
LGTM for the logic,

Just a couple of comments on the code style.

We can also use the functional options pattern, having in the end something more ideomatic like

`NewFactValue(thingstocast, WithSnakeCase(), WithStringCasting())` or `NewFactValue(thingstocast)` without any additional configuration, having the options as variadic arguments.

But this is up to you, we have two options for now, so as you wish.

The only things I request to change are in the comments.

Some examples of functional options pattern: https://charlesxu.io/go-opts/

pkg/factsengine/entities/fact_value.go
arbulu89: Initially I didn't want to do that because:
- This is "somehow" a public API, as it it is in the `package` folder, and it could be maybe used by some external user (on a plugin?). Anyway, i really doubt this
- It would force me to change the current usages of the code.

None of them a big deal. I will change.
Just writing down the rationale about this implementation

pkg/factsengine/entities/fact_value.go
arbulu89: I'm letting this as it is by now, to avoid any (even really not harmful) breaking change

pkg/factsengine/entities/fact_value.go
arbulu89: I have finally used the functional options pattern, which allows us to have the same signature.
It affect minimally to the usage, as we only have one default option by now (the StringConversion)

CDimonaco (approved):
LGTM!

pkg/factsengine/entities/fact_value.go
```diff
@@ -19,6 +20,27 @@ var ValueNotFoundError = FactGatheringError{
 	Message: "error getting value",
 }
 
+type conf struct {
+	snakeCaseKeys    bool
+	stringConversion bool
+}
+
+type FactValuOption func(f *conf)
```
rtorrero, line 28: Did you mean `FactValueOption` here?

rtorrero (approved):
LGTM

PR #263, opened 2023-09-26
Handle nil value in NewFactValue

arbulu89:
Handle `nil` value in `NewFactValue`. For instance, the new saptune gatherer can have some null values, so we need to send them as null.
This doesn't mean that in many gatherers, if we get kind of `null` value we don't need to send a fact gathering error.

Fact output example. The `()` is the null value in rhai:
```
#{
  "$schema": "file:///usr/share/saptune/schemas/1.0/saptune_solution_verify.schema.json",
  "argv": "saptune --format json solution verify",
  "command": "solution verify",
  "exit_code": 0,
  "messages": [],
  "pid": 11179,
  "publish_time": "2023-09-26 16:02:18.474",
  "result": #{
    "attentions": [],
    "notes_enabled": [],
    "system_compliance": (),
    "verifications": []
  }
} 
```

# How it was tested

I tested together with wanda and web.
The value arrives properly what we visualize in the frontend as `NULL_VALUE`. I don't know if we need to change something there:

image

### DIFF
--- a/pkg/factsengine/entities/fact_value.go
+++ b/pkg/factsengine/entities/fact_value.go
@@ -89,9 +89,20 @@ func NewFactValue(factInterface interface{}, opts ...FactValueOption) (FactValue
 			return ParseStringToFactValue(value), nil
 		}
 		return &FactValueString{Value: value}, nil
+	case nil:
+		return &FactValueNil{}, nil
 	default:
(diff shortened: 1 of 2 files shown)


rtorrero (approved):
LGTM @arbulu89, thanks!

EMaksy (commented):
Not an golang expert, but lgtm :D

PR #265, opened 2023-09-28
Fstab gatherer

CDimonaco:
This gatherer reads and parse the fstab file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,11 +27,15 @@ require (
 	google.golang.org/protobuf v1.31.0
 )
 
-require github.com/tredoe/osutil v1.0.6
+require (
+	github.com/d-tux/go-fstab v0.0.0-20141204152952-eb4090f26517
+	github.com/iancoleman/strcase v0.2.0
+	github.com/tredoe/osutil v1.0.6
+	golang.org/x/mod v0.9.0
+)
 
 require (
 	github.com/chigopher/pathlib v1.0.0 // indirect
-	github.com/d-tux/go-fstab v0.0.0-20141204152952-eb4090f26517 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/fatih/color v1.13.0 // indirect
 	github.com/fsnotify/fsnotify v1.6.0 // indirect
@@ -40,7 +44,6 @@ require (
 	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/hashicorp/hcl v1.0.0 // indirect
 	github.com/hashicorp/yamux v0.0.0-20180604194846-3520598351bb // indirect
-	github.com/iancoleman/strcase v0.2.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jinzhu/copier v0.3.5 // indirect
 	github.com/magiconair/properties v1.8.7 // indirect
@@ -59,7 +62,6 @@ require (
 	github.com/subosito/gotenv v1.4.2 // indirect
 	github.com/tklauser/numcpus v0.4.0 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
-	golang.org/x/mod v0.9.0 // indirect
 	golang.org/x/net v0.10.0 // indirect
 	golang.org/x/sys v0.8.0 // indirect
 	golang.org/x/term v0.8.0 // indirect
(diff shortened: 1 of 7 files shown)


go.mod
```diff
@@ -31,6 +31,7 @@ require github.com/tredoe/osutil v1.0.6
 
 require (
 	github.com/chigopher/pathlib v1.0.0 // indirect
+	github.com/d-tux/go-fstab v0.0.0-20141204152952-eb4090f26517 // indirect
```
arbulu89, line 34: Just golang  curiosity, why is this `indirect`?

internal/factsengine/gatherers/fstab.go
```diff
@@ -0,0 +1,100 @@
...
+	FstabFileDecodingError = entities.FactGatheringError{
+		Type:    "fstab-decoding-error",
+		Message: "error decoding fstab file",
+	}
+)
+
+type FstabEntry struct {
+	Device     string   `json:"device"`
+	MountPoint string   `json:"mount_point"`
+	FS         string   `json:"fs"`
```
arbulu89, line 33: I would rename this to `FileSystemType` or `Type` (`file_system_type/type`).
It makes it more obvious

internal/factsengine/gatherers/fstab.go
```diff
@@ -0,0 +1,100 @@
...
+
+func NewFstabGatherer(filePath string) *FstabGatherer {
+	return &FstabGatherer{fstabFilePath: filePath}
+}
+
+func NewDefaultFstabGatherer() *FstabGatherer {
+	return &FstabGatherer{fstabFilePath: FstabFilePath}
+}
+
+func (g *FstabGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
```
arbulu89, line 51: Maybe update this `g` by `f`

internal/factsengine/gatherers/fstab.go
```diff
@@ -0,0 +1,100 @@
...
+		return nil, err
+	}
+
+	var unmarshalled []interface{}
+	err = json.Unmarshal(marshalled, &unmarshalled)
+	if err != nil {
+		return nil, err
+	}
+
+	return entities.NewFactValue(unmarshalled, entities.WithStringConversion())
```
arbulu89, line 99: We don't need the `WithStringConversion`

test/fixtures/gatherers/fstab.valid
```diff
@@ -0,0 +1,8 @@
+# test fstab
+
+/dev/system/root  /                       btrfs  defaults                      0  0
+/dev/system/root  /root                   btrfs  subvol=/@/root                1  1
+/dev/system/root  /home                   btrfs  subvol=/@/home                1  0
+/dev/system/swap  swap                    swap   defaults                      0  0
+/dev/system/root  /.snapshots             btrfs  subvol=/@/.snapshots          0  1
+UUID=DA2F-21CF    /boot/efi               vfat   utf8
```
arbulu89, line 8: Is this correct? The last entry doesn't have the `Backup` and `Order`?
I see that in the tests they are set to `0` both of them

arbulu89 (changes requested):
Overall great!
You forgot to add the new gatherer in the `gatherers.go` file.

test/fixtures/gatherers/fstab.valid
arbulu89: Auto-answer: Yes, they are optional, and they default to 0 if they are not present:
https://www.man7.org/linux/man-pages/man5/fstab.5.html

go.mod
CDimonaco: An indirect dependency is a dependency added but not used, I added the dependency via go get, and I was not using before I wrote the code.

After you wrote the code the dependency is used and it's not indirect anymore, but go build and go run does not update the go mod file.

You should call directly go mod tidy to fix this type of inconsistency in the go mod, I've done that, thanks for spotting! :D

internal/factsengine/gatherers/fstab.go
CDimonaco: done

arbulu89 (approved):
Greeeat!

internal/factsengine/gatherers/fstab.go
CDimonaco: g was "gatherer", but I renamed to f, no hard feelings on that :D

internal/factsengine/gatherers/fstab.go
CDimonaco: done, damn c&p!

PR #268, opened 2023-09-29
sysctl gatherer

rtorrero:
This PR adds a `sysctl`. Note: Right now, it will fail if no argument is provided but if we want, we could allow zero arguments to return the whole sysctl map.


internal/factsengine/gatherers/sysctl.go
```diff
@@ -0,0 +1,118 @@
...
+
+func NewSysctlGatherer(executor utils.CommandExecutor) *SysctlGatherer {
+	return &SysctlGatherer{
+		executor: executor,
+	}
+}
+
+func (s *SysctlGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting sysctl facts gathering process")
```
arbulu89, line 49: We could use the string interpolation with `SysctlGathererName` as we do with many others.
Small change, but it would help us if we for some reason change the gatherer name, to not forget this details

internal/factsengine/gatherers/sysctl.go
```diff
@@ -0,0 +1,118 @@
...
+		} else {
+			gatheringError := SysctlValueNotFound.Wrap(factReq.Argument)
+			log.Error(gatheringError)
+			fact = entities.NewFactGatheredWithError(factReq, gatheringError)
+		}
+
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested %s facts gathered", CorosyncCmapCtlGathererName)
```
arbulu89, line 74: Forgot to change this I guess hehe

internal/factsengine/gatherers/sysctl.go
```diff
@@ -0,0 +1,118 @@
...
+	return facts, nil
+}
+
+func sysctlOutputToMap(output []byte) *entities.FactValueMap {
```
arbulu89, line 78: Could we try to use `envparse.Parse(output)` maybe?
At the end the output looks like an `.env` file format.
I know that this wouldn't create a map of maps, but I don't think we would need such a thing in this case. I expect the user putting the whole needed line to get individual values

I would try to change, you have the tests there to confirm that everything went right

PD: I know that this is based in the `corosynccmapctl` gatherer, but i think we could make it simpler this way

internal/factsengine/gatherers/sysctl_test.go
```diff
@@ -0,0 +1,184 @@
...
+			Argument: "kernel.domainname",
+		},
+	}
+
+	factResults, err := c.Gather(factRequests)
+
+	expectedResults := []entities.Fact{
+		{
+			Name:  "empty_value",
+			Value: &entities.FactValueString{Value: ""}, // We should probably use the upcoming FactValueNil (PR#263)
```
arbulu89, line 178: Up to you. No hard opinions.
Remember to document the way you use in both cases

arbulu89 (changes requested):
We need to add the gatherer to the list of standard gatherers
The `envparse` usage comment is a personal opinion. I think that it would simplify the code. Specially, because i don't expect the gatherer being called with "intermediate" values 

PD: Please, test things manually as well, to see that it does what we expect

internal/factsengine/gatherers/sysctl.go
rtorrero: Indeed it's based on the corosync one, I saw a comment in confluence mentioning that we might want to expose `sysctl -a` (with no arguments) as a map, so I used the same approach. In the end apparently we don't need this but if we do in the future, the current approach would be more powerful. In any case, no hard opinions on this.

arbulu89: Anyway, do we need all of this `switch` code with the `SplitN`?
I think the `corosynccmapctl` code is cleaner, and does exactly the same (99% the same)

```
parts := strings.Split(line, "=")
key := parts[0]
value := parts[1]
```

internal/factsengine/gatherers/sysctl.go
```diff
@@ -0,0 +1,107 @@
...
+			continue
+		}
+
+		key := strings.TrimSpace(parts[0])
+		value := strings.TrimSpace(parts[1])
+
+		cursor := outputMap
+		pathComponents := strings.Split(key, ".")
+
+		for i, component := range pathComponents {
```
arbulu89, line 93: I still think that the code in `corosynccmapctl` is simpler, with less `if/else` branching.
Anyway, I guess it works fine

internal/factsengine/gatherers/sysctl_test.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	factResults, err := c.Gather(factRequests)
+
+	suite.EqualError(err, expectedError.Error())
+
+	suite.Empty(factResults)
+}
+
+func (suite *SysctlTestSuite) TestSysctlGatherer() {
```
arbulu89, line 130: As we are going with the "map" version, could we add a more complex scenario?
Like asking for `user` and getting all the children values

internal/factsengine/gatherers/sysctl.go
```diff
@@ -0,0 +1,107 @@
...
+	return facts, nil
+}
+
+func sysctlOutputToMap(output []byte) *entities.FactValueMap {
+	outputMap := &entities.FactValueMap{Value: make(map[string]entities.FactValue)}
+
+	for _, line := range strings.Split(string(output), "\n") {
+		parts := strings.SplitN(line, "=", 2)
+		if len(line) == 0 || len(parts) != 2 {
+			continue
```
arbulu89, line 84: I don't know if we shouldn't raise an error in this case, as we would get a malformed output

arbulu89 (approved):
Green light from my side.
Added some comments, but judge yourself :D

internal/factsengine/gatherers/sysctl_test.go
arbulu89: You could've added the request in this very same test, but well, a new test is always ok

internal/factsengine/gatherers/sysctl.go
arbulu89: Some sort of `decoding` error as we do in other gatherers

CDimonaco (approved):

PR #269, opened 2023-09-29
Dir scan gatherer

CDimonaco:
This PR implements `dir-scan` gatherer, this gatherer will be used in a variety of checks.

Given a glob pattern as argument, the gatherer returns a map with directories as keys (directories are matched with the glob pattern) and for value a map containing the group id, user id of the directory, and the files of the directory. The files will also matched by the glob pattern as argument.


internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func TestDirScanGathererSuite(t *testing.T) {
+	suite.Run(t, new(DirScanGathererSuite))
+}
+
+func (s *DirScanGathererSuite) SetupSuite() {
+	tFs := afero.NewMemMapFs()
+
+	for i := 0; i <= 2; i++ {
```
arbulu89, line 28: Only for 3 entries, it would be even simpler to have them in the same way you create the "bad" files hehe
It is easier to visually see the created folders

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,154 @@
...
+	s.testFS = tFs
+}
+
+func (s *DirScanGathererSuite) TestDirScanningErrorDirScaningWithoutGlob() {
+	g := gatherers.NewDirScanGatherer(s.testFS)
+
+	fr := []entities.FactRequest{{
+		Argument: fmt.Sprintf("%s/1/ASCS3", dirScanTestBasePath),
+		CheckID:  "check1",
+		Gatherer: "dir-scan",
```
arbulu89, line 46: Remember to change this names with `-` if you change the gatherer name

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	_, _ = tFs.Create(fmt.Sprintf("%s/1/ASCS3", dirScanTestBasePath))
+	_, _ = tFs.Create(fmt.Sprintf("%s/1/ASDX2", dirScanTestBasePath))
+	_, _ = tFs.Create(fmt.Sprintf("%s/2/ASDX1", dirScanTestBasePath))
+
+	s.testFS = tFs
+}
+
+func (s *DirScanGathererSuite) TestDirScanningErrorDirScaningWithoutGlob() {
```
arbulu89, line 40: Why `Error` in the name?

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,154 @@
...
+		Name:     "dir-scan",
+	}}
+
+	result, _ := g.Gather(fr)
+	expectedResult := []entities.Fact{{
+		Name:    "dir-scan",
+		CheckID: "check1",
+		Value: &entities.FactValueMap{
+			Value: map[string]entities.FactValue{
+				"/var/test/1": &entities.FactValueMap{
```
arbulu89, line 56: I'm a bit surprised with this.
If I give `/var/test/1/ASCS3` as argument, I would expect to get the owner:group of this file/folder.
And if it is a folder, the "children" files

arbulu89, line 56: I'm a bit surprised with this.
If I give `/var/test/1/ASCS3` as argument, I would expect to get the owner:group of this file/folder.
And if it is a folder, the "children" files

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,154 @@
...
+		Gatherer: "dir-scan",
+		Name:     "dir-scan",
+	}}
+
+	expectedResult := []entities.Fact{{
+		Name:    "dir-scan",
+		CheckID: "check1",
+		Value: &entities.FactValueMap{
+			Value: map[string]entities.FactValue{
+				"/var/test/0": &entities.FactValueMap{
```
arbulu89, line 113: I'm lost again with this output

arbulu89 (commented):
Hey @CDimonaco 
I'm lost on the implementation. I just followed the tests and it is not what I expected at least (or I don't get it).

If I use the `"/var/test/*/ASCS*"` wildcar, I would expect to get as entries all the files/folders that match completely this path (/var/test/1/ASCS00 and /var/test/2/ASCS01 for example). And the owner:group + subfiles


PD: Add the gatherer to the list of standard gatherers

internal/factsengine/gatherers/dir_scan.go
```diff
@@ -0,0 +1,152 @@
...
+	"syscall"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+
+	"github.com/trento-project/agent/pkg/factsengine/entities"
+)
+
+const (
+	DirScanGathererName = "dir-scan"
```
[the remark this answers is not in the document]
CDimonaco: done

CDimonaco:
@arbulu89 Adjusted output, now the folder files have group/user information like directories.

In the test I had to switch to a normal fs, using always afero to create a namespaced temp fs directory, the cleanup and creation are done in the test suite.

A normal fs is needed because we need to access to `syscall.Stat_t` which is available only on "real" fs, so afero MemFS was not suitable for this type of test.

A solution could be to add another interface, but the cost is not justified, using always afero but with a temp directory based fs is the most reasonable, correct, approach. We can test also the casting to Stat_t error handling etc..

I documented this change also in the corresponding commit.

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,207 @@
...
+	suite.Run(t, new(DirScanGathererSuite))
+}
+
+func (s *DirScanGathererSuite) SetupSuite() {
+	bfs := afero.NewOsFs()
+
+	s.basePathFS = afero.GetTempDir(bfs, "")
+	tFs := afero.NewBasePathFs(bfs, s.basePathFS)
+	for i := 0; i <= 2; i++ {
+		dirPath := fmt.Sprintf("%s/%d/", dirScanTestBasePath, i)
```
arbulu89, line 34: `filepath.Join` looks more appropriate to create the file paths, instead of `Sprintf`

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,207 @@
...
+		_ = tFs.MkdirAll(dirPath, 0777)
+		_, _ = tFs.Create(filePath)
+	}
+
+	_, _ = tFs.Create(fmt.Sprintf("%s/1/ASCS3", dirScanTestBasePath))
+	_, _ = tFs.Create(fmt.Sprintf("%s/1/ASDX2", dirScanTestBasePath))
+	_, _ = tFs.Create(fmt.Sprintf("%s/2/ASDX1", dirScanTestBasePath))
+
+	s.testFS = tFs
+
```
arbulu89, line 45: Could we remove this new line after `s.testFs` and add a new one before `func`

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,207 @@
...
+func (s *DirScanGathererSuite) TestDirScanningSuccess() {
+	groupSearcher := mocks.NewGroupSearcher(s.T())
+	groupSearcher.On("GetGroupByID", mock.AnythingOfType("string")).Return("trento", nil)
+
+	userSearcher := mocks.NewUserSearcher(s.T())
+	userSearcher.On("GetUsernameByID", mock.AnythingOfType("string")).Return("trento", nil)
+
+	g := gatherers.NewDirScanGatherer(s.testFS, userSearcher, groupSearcher)
+	fr := []entities.FactRequest{{
+		Argument: dirScanTestGlobPattern,
```
arbulu89, line 131: Is it not better to put this `dirScanTestGlobPattern` here locally in the test?
It is not used anywhere else, and it makes the reading of the test more difficult, as you need to scroll up and down

internal/factsengine/gatherers/dir_scan.go
```diff
@@ -0,0 +1,194 @@
...
+	stat, ok := fi.Sys().(*syscall.Stat_t) //nolint
+	if !ok {
+		return nil, fmt.Errorf("could not extract stat infos for file %s", path)
+	}
+	uid := strconv.Itoa(int(stat.Uid))
+	gid := strconv.Itoa(int(stat.Gid))
+
+	group, err := d.groupSearcher.GetGroupByID(gid)
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve group for gigroupd %s", gid)
```
arbulu89, line 164: typo on `gigroupd`? maybe it should be `gid`

internal/factsengine/gatherers/dir_scan.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	user, err := d.userSearcher.GetUsernameByID(uid)
+	if err != nil {
+		return nil, fmt.Errorf("could not retrieve group for uid %s", uid)
```
arbulu89, line 168: typo -> `could not retrieve usename for uid`

internal/factsengine/gatherers/dir_scan.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return &DirScanStatInfo{
+		Group: group,
+		Owner: user,
+	}, nil
+}
+
+func getDirScanResultKeyFromPath(path string) string {
```
arbulu89, line 177: Is this function reason to exist just "informative"?

arbulu89 (commented):
The core code looks great.
I'm still dubious in one thing. Whether we should have the parent dir or not.

For example:
If I use the pattern `/etc/polkit-1/rules.d/[0-9][0-9]-SAP[A-Z][A-Z0-9][A-Z0-9]-[0-9][0-9].rules`.
In this case, what I would personally like is to get the files/folders that match this glob, and not maybe the parent.

The same for `/usr/sap/[A-Z][A-Z0-9][A-Z0-9]/ASCS[0-9][0-9]`. I would like to get the files that match this files.

So I would have the result as:
```
[
#{
file: /usr/sap/NWP/ASCS00,
owner: trento,
group: trento
},
#{
file: /usr/sap/QAS/ASCS00,
owner: trento,
group: trento
},
]
```
At the end I don't see much value on having the parent folder there. Just with the children we would achieve all we want. If somebody wants the parent folder owner/group, i guess they can gather again with other glob.

Maybe the gatherer name is not the most accurate one. Maybe `file_scan` would be better.
We can discuss this offline if you want.

@abravosuse @rtorrero What do you think?

PD: I left some cosmetic/typo kind of comments

internal/factsengine/gatherers/dir_scan.go
```diff
@@ -0,0 +1,194 @@
...
+				Owner: statInfo.Owner,
+				Group: statInfo.Group,
+				Files: nil,
+			})
+			result[resultKey] = resultEntry
+			continue
+		}
+
+		newEntry := DirScanDetails{
+			Owner: statInfo.Owner,
```
arbulu89, line 132: In fact, after testing in a real env, this approach looks wrong.
Here, we are creating a new "parent" folder with owner/group of the child. So the owner/group is inherited from the first file that matches the glob. Meaning that the parent folder doesn't have its own stats.
Anyway, this wouldn't have any impact if we decide to remove this "parent" folder entry

abravosuse:
@arbulu89 not sure to follow because I cannot see what @CDimonaco code does.

Let's say the global pattern is `/usr/sap/[A-Z][A-Z0-9][A-Z0-9]/ASCS[0-9][0-9]`. Then I don't expect to see `/usr/sap/NWP` in the output. 

In order to get `/usr/sap/NWP`, the glob patter would have to be  `/usr/sap/[A-Z][A-Z0-9][A-Z0-9]`.

If that's what you mean, then I agree with you.

internal/factsengine/gatherers/dir_scan.go
```diff
@@ -0,0 +1,194 @@
...
+	return NewDirScanGatherer(afero.NewOsFs(), &cf, &cf)
+}
+
+func (d *DirScanGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	log.Infof("Starting %s facts gathering process", DirScanGathererName)
+	facts := []entities.Fact{}
+
+	for _, requestedFact := range factsRequests {
+		if requestedFact.Argument == "" {
+			facts = append(facts, entities.NewFactGatheredWithError(requestedFact, &DirScanMissingArgumentError))
```
rtorrero, line 83: I'd add a `log.Error(DirScanMissingArgumentError.Message)` here, might be helpful when diagnosing issues

rtorrero (commented):
appart from the comments from @arbulu89 and the small comment I left in the code, was expecting that it would also handle paths with a trailing `/` in the end, but it seems is not the case. @arbulu89 has told me you are agreed on additional changes that should fix this, so it shouldn't be a problem.

Good stuff :-)

internal/factsengine/gatherers/dir_scan.go
CDimonaco: done

internal/factsengine/gatherers/dir_scan.go
CDimonaco: done

internal/factsengine/gatherers/dir_scan_test.go
CDimonaco: you are right but due to the nature of paths, interpolations, convertion to string from int for the index, it's pretty deterministic and it's a test so I used sprintf on purpose to avoid other boilerplate code

internal/factsengine/gatherers/dir_scan_test.go
CDimonaco: done

CDimonaco:
Changed to a file list approach, addressed all feedbacks, thanks to everyone!

internal/factsengine/gatherers/dir_scan_test.go
arbulu89: Still `Error` in the test name, which is a normal test, without raising any error

internal/factsengine/gatherers/dir_scan_test.go
```diff
@@ -0,0 +1,168 @@
...
+						"group": &entities.FactValueString{Value: "trento"},
+						"name":  &entities.FactValueString{Value: "/var/test/1/ASCS3"},
+					},
+				},
+			},
+		},
+	}}
+
+	s.EqualValues(expectedResult, result)
+
```
arbulu89, line 85: We could remove this empty line

arbulu89 (approved):
Looks great!
I even tested in a real env, and the `/usr/sap/[A-Z][A-Z0-9][A-Z0-9]/ERS[0-9][0-9]` kind of patterns work great.

Don't forget adding the docs in wanda

PR #274, opened 2023-10-10
Resolver gatherer

rtorrero:
This PR adds a resolver gatherer, it uses the filesystem to search for running sap systems and the profile data to get the virtual hostnames associated to each instance of the sap system. It will then attempt to resolve those hostnames to confirm that the OS knows about them.


internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	sysNames := GetSIDsString(systems)
+	return sysNames, nil
+}
+
+func (r *SapLocalhostResolverGatherer) getProfiles() ([]map[string]string, error) {
```
arbulu89, line 112: I think we could simplify this by (pseudocode):
``` 
func ... getInstanceHostnames([]string, error) {
	systems, err := sapsystem.FindSystems(r.fs)

	if err != nil {
		return nil, err
	}

        for _, system := range systems {
		sid := filepath.Base(system)

                profileFiles, err: = FindProfiles(sid) {
                     hostname, err := applyRegexPatternToFindHostnameInProfileFileName...
                     hostnames = append(hostnames, hostname)
                }
        }

        return hostnames, nil

}
```

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+	return &SapLocalhostResolverGatherer{fs: fs, hr: hr}
+}
+
+func (r *SapLocalhostResolverGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	var facts []entities.Fact
+	profiles, _ := r.getProfiles()
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+		for _, profile := range profiles {
```
arbulu89, line 45: We iterate here over hostnames better using the `getInstanceHostnames` function

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+
+				// reachable := false
+				// for _, ip := range ips {
+				// 	if ok, _ := isHostReachable(ip); ok {
+				// 		reachable = true
+				// 		break
+				// 	}
+				// }
+
+				factValue := &entities.FactValueMap{
```
arbulu89, line 65: You don't really need to create the `FactValueMap` yourself.
You can create your own internal struct with its own json annotations, and later on marshall/unmarshall/convertToFactValue as we do in many other gatherers

arbulu89 (commented):
@rtorrero Some initial comments about the generic logic of the code

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -28,6 +35,12 @@ type SapLocalhostResolverGatherer struct {
 	hr utils.HostnameResolver
 }
 
+type ReachabilityDetails struct {
```
arbulu89, line 38: Wouldn't be `ResorvabilityDetails` better? (what a name :laughing: )

internal/factsengine/gatherers/gatherer.go
```diff
@@ -10,20 +10,21 @@ type FactGatherer interface {
...
-		SBDDumpGathererName:         NewDefaultSBDDumpGatherer(),
-		SystemDGathererName:         NewDefaultSystemDGatherer(),
-		VerifyPasswordGathererName:  NewDefaultPasswordGatherer(),
+		CibAdminGathererName:             NewDefaultCibAdminGatherer(),
+		CorosyncCmapCtlGathererName:      NewDefaultCorosyncCmapctlGatherer(),
+		CorosyncConfGathererName:         NewDefaultCorosyncConfGatherer(),
+		FstabGathererName:                NewDefaultFstabGatherer(),
+		GroupsGathererName:               NewDefaultGroupsGatherer(),
+		HostsFileGathererName:            NewDefaultHostsFileGatherer(),
+		SapLocalhostResolverGathererName: NewDefaultSapLocalhostResolverGatherer(),
```
arbulu89, line 19: Alphabetical order please

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,123 @@
...
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/trento-project/agent/internal/core/sapsystem"
+	"github.com/trento-project/agent/pkg/factsengine/entities"
+	"github.com/trento-project/agent/pkg/utils"
+)
+
+const (
+	SapLocalhostResolverGathererName = "saplocalhost_resolver"
```
arbulu89, line 16: Let's decide the naming after we have the code hehe

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,123 @@
...
+		fact = entities.NewFactGatheredWithRequest(factReq, factValue)
+		facts = append(facts, fact)
+	}
+
+	return facts, nil
+}
+
+func (r *SapLocalhostResolverGatherer) getInstanceHostnameDetails() (map[string]ReachabilityDetails, error) {
+	systems, err := sapsystem.FindSystems(r.fs)
+	reachabilityDetails := make(map[string]ReachabilityDetails)
```
arbulu89, line 73: You could initialized this after the next error checking, as it is not needed until that point

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,123 @@
...
+		sid := filepath.Base(system)
+		profileFiles, err := sapsystem.FindProfiles(r.fs, sid)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, profileFile := range profileFiles {
+			result := make(map[string]string)
+			match := hostnameRegexCompiled.FindStringSubmatch(profileFile)
+			if len(match) == 0 {
```
arbulu89, line 89: It wouldn't be better doing `if len(match) != 4 { continue }` and using the `match[1]`, `match[2]`, etc
I know that the `SubexpNames` looks useful, but it makes the code more complex in my opinion, as it includes the need for a new map variable and a `for+if` loops

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,123 @@
...
+				if i != 0 && name != "" {
+					result[name] = match[i]
+				}
+			}
+			details := ReachabilityDetails{
+				Hostname:     result["Hostname"],
+				Addresses:    []string{},
+				InstanceName: result["InstanceName"],
+			}
+			details.Addresses, _ = r.hr.LookupHost(details.Hostname)
```
arbulu89, line 102: I think it would be better to run this before creating the details (check the error), and compose the details as:
```
addresses, err = r.hr.LookupHost(details.Hostname)
if err != nil {
    return nil, err
}
details := eachabilityDetails{
    Hostname: match[3]
    Addresses:    addresses,
    InstanceName: match[2],
}
```

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,123 @@
...
+			reachabilityDetails[result["SID"]] = details
```
arbulu89, line 103: I guess this is `wip` yet

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,123 @@
...
+
+	rd, err := r.getInstanceHostnameDetails()
+	if err != nil {
+		log.Error(err)
+		return nil, SapLocalhostResolverHostnameResolutionError.Wrap(err.Error())
+	}
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+		factValue, _ := mapReachabilityDetailsToFactValue(rd)
```
arbulu89, line 63: check for errors please

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (suite *SapLocalhostResolverTestSuite) TestSapLocalhostResolverSuccess() {
+	appFS := afero.NewMemMapFs()
+
+	err := appFS.MkdirAll("/usr/sap/PRD", 0644)
+	suite.NoError(err)
+	err = appFS.MkdirAll("/usr/sap/QAS", 0644)
+	suite.NoError(err)
+
+	defaultProfileFile, _ := os.Open(helpers.GetFixturePath("gatherers/sap_profile.default"))
```
arbulu89, line 37: No need for this. You can create the files with `[]byte{}` as you don't care the content

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,95 @@
...
+	defaultProfileContent, _ := io.ReadAll(defaultProfileFile)
+	err = afero.WriteFile(appFS, "/sapmnt/PRD/profile/DEFAULT.PFL", defaultProfileContent, 0644)
+	suite.NoError(err)
+	err = afero.WriteFile(appFS, "/sapmnt/PRD/profile/DEFAULT.1.PFL", []byte{}, 0644)
```
arbulu89, line 41: We don't need this now. The `FindProfiles` is already tested and we know that it excludes the backups

arbulu89 (commented):
Looking good @rtorrero 
Good progress

pkg/utils/hostnameresolver.go
```diff
@@ -0,0 +1,15 @@
+package utils
+
+import "net"
+
+//go:generate mockery --name=HostnameResolver
+
+type HostnameResolver interface {
```
arbulu89, line 7: Maybe we don't need a new file for this. This interface is only used by this gatherer, so we could have it directly in the gatherer file.
We already to this with other gatherers: https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/systemd.go#L34

And you pass the `Resolver` or `DefaultResolver` from golang net package:
https://cs.opensource.google/go/go/+/master:src/net/lookup.go;l=128

`net.DefaultResolver{}` and use it as `r.LookupHost(context.Background(), host)`

Anyway, not a big deal

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+
+	for _, system := range systems {
+		sid := filepath.Base(system)
+		profileFiles, err := sapsystem.FindProfiles(r.fs, sid)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, profileFile := range profileFiles {
+			if filepath.Base(profileFile) == "DEFAULT.PFL" {
```
arbulu89, line 108: why `filepath.Base`? The profile file is `DEFAULT.PFL` already

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		facts = append(facts, fact)
+	}
+
+	return facts, nil
+}
+
+func (r *SapLocalhostResolverGatherer) getInstanceHostnameDetails() (map[string]ResolvabilityDetails, error) {
+	systems, err := sapsystem.FindSystems(r.fs)
+	reachabilityDetails := make(map[string]ResolvabilityDetails)
```
arbulu89, line 94: This should be `make(map[string][]ResolvabilityDetails)`. We need a list of `ResolvabilityDetails`, as we check for each hostname.

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			details := ResolvabilityDetails{
+				Hostname:     match[3],
+				Addresses:    addresses,
+				InstanceName: match[2],
+			}
+			reachabilityDetails[match[1]] = details
```
arbulu89, line 126:
```
if detailsList, ok := reachabilityDetails[match[1]]; !ok {
    reachabilityDetails[match[1]] = []ResolvabilityDetails{details}
} else {
    reachabilityDetails[match[1]] = append(reachabilityDetails[match[1]], details)
}
```
or something like that, otherwise you are overwritting all the time

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return reachabilityDetails, nil
+}
+
+func mapReachabilityDetailsToFactValue(entries map[string]ResolvabilityDetails) (entities.FactValue, error) {
```
arbulu89, line 133: `map[string][]ResolvabilityDetails`

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func NewSapLocalhostResolver(fs afero.Fs, hr utils.HostnameResolver) *SapLocalhostResolverGatherer {
+	return &SapLocalhostResolverGatherer{fs: fs, hr: hr}
+}
+
+func (r *SapLocalhostResolverGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	var facts []entities.Fact
+
+	rd, err := r.getInstanceHostnameDetails()
```
arbulu89, line 71: `rd` ? maybe better `details`

arbulu89 (commented):
More comments

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,154 @@
...
+func NewDefaultSapLocalhostResolverGatherer() *SapLocalhostResolverGatherer {
+	return NewSapLocalhostResolver(afero.NewOsFs(), utils.Resolver{})
+}
+
+func NewSapLocalhostResolver(fs afero.Fs, hr utils.HostnameResolver) *SapLocalhostResolverGatherer {
+	return &SapLocalhostResolverGatherer{fs: fs, hr: hr}
+}
+
+func (r *SapLocalhostResolverGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := make([]entities.Fact, 0, len(factsRequests))
```
CDimonaco, line 74: Please use a direct slice creation without make with a fixed capacity, the code looks less readable and I think we don't need such a thing in this context performance wise

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,154 @@
...
+			if err != nil {
+				return nil, SapLocalhostResolverHostnameResolutionError.Wrap(err.Error())
+			}
+
+			details := ResolvabilityDetails{
+				Hostname:     match[3],
+				Addresses:    addresses,
+				InstanceName: match[2],
+			}
+			if _, ok := resolvabilityDetails[match[1]]; !ok {
```
CDimonaco, line 130: Could you use found, instead of ok? Ok is used more in typecasting scenarios, found is more appropriate if you are checking if you found something in the map

pkg/utils/hostnameresolver.go
CDimonaco: I completely agree, and anyway I think it's not the place in package utils, If you want to put in a different file you can have it in the gatherers package, like we did with groups and users lookup for the dir-scan gatherer

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/trento-project/agent/pkg/utils"
+)
+
+const (
+	SapLocalhostResolverGathererName = "saplocalhost_resolver"
+)
+
+// nolint:gochecknoglobals
+var (
+	subgroupMapping = map[int]string{
```
CDimonaco, line 22: Could we squash this in the sprintf below? Wdyt? Completely up to you, We use is with len in other code, maybe a map is not the appropriate structure?

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,154 @@
...
+			return nil, SapLocalhostResolverFileSystemError.Wrap(err.Error())
+		}
+
+		for _, profileFile := range profileFiles {
+			if profileFile == "DEFAULT.PFL" {
+				continue
+			}
+
+			match := hostnameRegexCompiled.FindStringSubmatch(profileFile)
+			if len(match) != len(subgroupMapping)+1 {
```
CDimonaco, line 117: The len thing, as commented above

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,154 @@
...
+				continue
+			}
+			addresses, err := r.hr.LookupHost(match[3])
+			if err != nil {
+				return nil, SapLocalhostResolverHostnameResolutionError.Wrap(err.Error())
+			}
+
+			details := ResolvabilityDetails{
+				Hostname:     match[3],
```
CDimonaco, line 126: Could we assign the match results to a local variables so it's more readable, instead of directly accessing with index every time?

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,154 @@
...
+	resolvabilityDetails := make(map[string][]ResolvabilityDetails)
+	for _, system := range systems {
+		sid := filepath.Base(system)
+		profileFiles, err := sapsystem.FindProfiles(r.fs, sid)
+		if err != nil {
+			return nil, SapLocalhostResolverFileSystemError.Wrap(err.Error())
+		}
+
+		for _, profileFile := range profileFiles {
+			if profileFile == "DEFAULT.PFL" {
```
CDimonaco, line 112: Could we use a constant here?

CDimonaco (changes requested):
I think the code looks good, I'm requesting changes not on the logic itself, but on style and readability

Anyway GJ

rtorrero:
@CDimonaco @arbulu89 thanks for the reviews, I think I've addressed everything, the only thing that I'm not 100% I did accordingly was the mocks, I didn't fully understand @arbulu89 proposal but I think I came up with a solution that addresses the main concern of not adding stuff to the utils when it's not being used elsewhere. In addition to the comments, I also:
 - Added the 'reachability' details (using ping) in the resulting fact value.
 - Improved a bit the main success test to include additional profiles and a failing ping.

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+
+	details, err := r.getInstanceHostnameDetails()
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	for _, factReq := range factsRequests {
+		var fact entities.Fact
+		factValue, err := mapReachabilityDetailsToFactValue(details)
```
arbulu89, line 102: We don't need to run `mapReachabilityDetailsToFactValue` within the loop. It uses always the same argument.
We should run it outside

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	resolvabilityDetails := make(map[string][]ResolvabilityDetails)
+	for _, system := range systems {
+		sid := filepath.Base(system)
+		profileFiles, err := sapsystem.FindProfiles(r.fs, sid)
+		if err != nil {
+			return nil, SapLocalhostResolverFileSystemError.Wrap(err.Error())
```
arbulu89, line 126: return `err` here

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+
+		for _, profileFile := range profileFiles {
+			if profileFile == sapsystem.SapDefaultProfile {
+				continue
+			}
+
+			match := hostnameRegexCompiled.FindStringSubmatch(profileFile)
+			if len(match) != regexSubgroupsCount {
+				continue
```
arbulu89, line 136: This `if len(match) != regexSubgroupsCount` is an exception, we should print some warning message I guess

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+			matchedSID := match[1]
+			matchedInstanceName := match[2]
+			matchedHostname := match[3]
+
+			addresses, err := r.hr.LookupHost(matchedHostname)
+			if err != nil {
+				return nil, SapLocalhostResolverHostnameResolutionError.Wrap(err.Error())
```
arbulu89, line 144: We shouldn't error in this case. We should set an empty list of addresses.
Otherwise, a failure in 1 hostname condemns the fact gathering to fail, while some other honstnames might be resolvable

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+
+			details := ResolvabilityDetails{
+				Hostname:     matchedHostname,
+				Addresses:    addresses,
+				InstanceName: matchedInstanceName,
+				Reachability: r.hp.Ping(matchedHostname),
+			}
+			if _, found := resolvabilityDetails[match[1]]; !found {
```
arbulu89, line 153: `match[1]` -> `matchedSID`.
And maybe a new line before the `if` to improve readability

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,154 @@
...
+import (
+	"errors"
+	"io"
+	"os"
+	"testing"
+
+	"github.com/spf13/afero"
+	"github.com/stretchr/testify/suite"
+	"github.com/trento-project/agent/internal/factsengine/gatherers"
+	mocks "github.com/trento-project/agent/internal/factsengine/gatherers/mocks"
```
arbulu89, line 12: No need for this `mocks` alias, you don't change name after all

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,154 @@
...
+	suite.EqualError(err, "fact gathering error: saplocalhost_resolver-file-system-error - error reading the sap profiles file system: open /sapmnt/QAS/profile: file does not exist")
+}
+
+func (suite *SapLocalhostResolverTestSuite) TestSapLocalhostResolverLookupHostError() {
+	appFS := afero.NewMemMapFs()
+
+	err := appFS.MkdirAll("/usr/sap/QAS", 0644)
+	suite.NoError(err)
+
+	ascsProfileFile, _ := os.Open(helpers.GetFixturePath("gatherers/sap_profile.ascs"))
```
arbulu89, line 135: I don't think we need to get the content of a file. We don't use it.
```
err = afero.WriteFile(appFS, "/sapmnt/QAS/profile/QAS_ASCS00_sapqasas", []byte{}, 0644)
```
works for us

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,154 @@
...
+	factRequests := []entities.FactRequest{{
+		Name:     "sap_localhost_resolver",
+		Gatherer: "sap_localhost_resolver",
+		CheckID:  "check1",
+	}}
+
+	factResults, err := g.Gather(factRequests)
+
+	suite.Nil(factResults)
+	suite.EqualError(err, "fact gathering error: saplocalhost_resolver-resolution-error - error resolving hostname: lookup sapqasas on 169.254.169.254:53: dial udp 169.254.169.254:53: connect: no route to host")
```
arbulu89, line 153: I find this outcome strange. Shouldn't we return an empty list of addresses here?
Otherwise, if we have 4 hostnames, and the resolution of one of them fails, the returned fact gathering would be failure, and maybe other hostnames resolve properly
PD: Commented in the prod code as well

test/fixtures/gatherers/sap_profile.er.minimal
```diff
@@ -0,0 +1,3 @@
+SAPSYSTEMNAME = NWP
```
arbulu89, line 1: I don't think we need this new file

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,154 @@
...
+	suite.mockPinger = new(mocks.HostPinger)
+}
+
+func (suite *SapLocalhostResolverTestSuite) TestSapLocalhostResolverSuccess() {
+	appFS := afero.NewMemMapFs()
+
+	err := appFS.MkdirAll("/usr/sap/QAS", 0644)
+	suite.NoError(err)
+
+	ascsProfileFile, _ := os.Open(helpers.GetFixturePath("gatherers/sap_profile.ascs"))
```
arbulu89, line 38: No need to get the file content

internal/factsengine/gatherers/saplocalhostresolver_test.go
```diff
@@ -0,0 +1,154 @@
...
+	ascsProfileConcent, _ := io.ReadAll(ascsProfileFile)
+	err = afero.WriteFile(appFS, "/sapmnt/QAS/profile/QAS_ASCS00_sapqasas", ascsProfileConcent, 0644)
+	suite.NoError(err)
+
+	erProfileFile, _ := os.Open(helpers.GetFixturePath("gatherers/sap_profile.er.minimal"))
+	erProfileConcent, _ := io.ReadAll(erProfileFile)
+	err = afero.WriteFile(appFS, "/sapmnt/QAS/profile/NWP_ER10_sapnwper", erProfileConcent, 0644)
```
arbulu89, line 45: This path is wrong. If should have `QAS` as `NWP` or viceversa. Not both. And better `ERS` than `ER`
By the way, I think we should have some test with a couple of SIDs and a couple of profiles

internal/factsengine/gatherers/mocks/HostPinger.go
```diff
@@ -0,0 +1,38 @@
+// Code generated by mockery v2.36.0. DO NOT EDIT.
```
arbulu89, line 1: Really strange, the 2 mocks are created with a different version of `mockery`

go.mod
```diff
@@ -14,28 +14,28 @@ require (
...
 	github.com/sirupsen/logrus v1.9.3
-	github.com/spf13/afero v1.9.5
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.16.0
 	github.com/stretchr/testify v1.8.4
 	github.com/tklauser/go-sysconf v0.3.10 // indirect
 	github.com/trento-project/contracts/go v0.0.0-20230823130307-95ed2147fa9d
-	github.com/vektra/mockery/v2 v2.32.3
+	github.com/vektra/mockery/v2 v2.36.0
```
arbulu89, line 23: Do we need this update?
If so, please, use the same mockery version to create the mocks.
(I would abstein for updating for the sake of updating, otherwise, all of use now need to update)

go.mod
```diff
@@ -14,28 +14,28 @@ require (
...
 	github.com/d-tux/go-fstab v0.0.0-20141204152952-eb4090f26517
 	github.com/iancoleman/strcase v0.2.0
+	github.com/spf13/afero v1.9.5
 	github.com/tredoe/osutil v1.0.6
 	golang.org/x/mod v0.9.0
 )
 
 require (
-	github.com/chigopher/pathlib v1.0.0 // indirect
+	github.com/chigopher/pathlib v0.15.0 // indirect
```
arbulu89, line 38: Going back in version?
Strange. Please have a look on this and try to get a clean update of the file

arbulu89 (changes requested):
Suggestions from my side

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+func (r *Resolver) LookupHost(host string) ([]string, error) {
+	return net.LookupHost(host)
+}
+
+func (p *Pinger) Ping(host string) bool {
+	pinger, err := probing.NewPinger(host)
+	if err != nil {
+		return false
+	}
+	pinger.Count = 3
```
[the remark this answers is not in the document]
rtorrero: it could create false positives, but after a bit of thought it's actually something that we would want to capture, so I'll bring it down to 1

internal/factsengine/gatherers/saplocalhostresolver.go
```diff
@@ -0,0 +1,177 @@
...
+func NewDefaultSapLocalhostResolverGatherer() *SapLocalhostResolverGatherer {
+	return NewSapLocalhostResolver(afero.NewOsFs(), &Resolver{}, &Pinger{})
+}
+
+func NewSapLocalhostResolver(fs afero.Fs, hr HostnameResolver, hp HostPinger) *SapLocalhostResolverGatherer {
+	return &SapLocalhostResolverGatherer{fs: fs, hr: hr, hp: hp}
+}
+
+func (r *SapLocalhostResolverGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := make([]entities.Fact, 0)
```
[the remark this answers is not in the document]
CDimonaco: Yes, this was my suggestion also, we don't need to preallocate anything

go.mod
rtorrero: Missed this non-requested change, I guess from `go mod`, removed (as well as the mockery one)

abravosuse:
@rtorrero @arbulu89 It's not the instance name that we are trying to resolve but the instance (virtual) hostname. Therefore my preference would be `SapInstanceHostnameResolver`.

internal/factsengine/gatherers/sapinstancehostnameresolver.go
```diff
@@ -14,19 +14,19 @@ import (
...
 // nolint:gochecknoglobals
 var (
-	hostnameRegexCompiled            = regexp.MustCompile(`(.+)_(.+)_(.+)`) // <SID>_<InstanceNumber>_<Hostname>
-	regexSubgroupsCount              = 4
-	SapLocalhostResolverDetailsError = entities.FactGatheringError{
-		Type:    "saplocalhost_resolver-details-error",
+	hostnameRegexCompiled                   = regexp.MustCompile(`(.+)_(.+)_(.+)`) // <SID>_<InstanceNumber>_<Hostname>
+	regexSubgroupsCount                     = 4
+	SapInstanceHostnameResolverDetailsError = entities.FactGatheringError{
+		Type:    "sapinstance_hostname_resolver-details-error",
```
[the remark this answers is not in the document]
rtorrero: no worries, I can change it

EMaksy (commented):
Hey nice work, lgtm :D

CDimonaco (approved):

PR #278, opened 2023-10-16
Systemd V2 gatherer

arbulu89:
SystemD V2 gatherer.
In this version, instead of returning the `active_state` value, it returns the next object:
```
type systemdUnitStatus struct {
	ActiveState      interface{} `json:"active_state"`
	Description      interface{} `json:"description"`
	ID               interface{} `json:"id"`
	LoadState        interface{} `json:"load_state"`
	NeedDaemonReload interface{} `json:"need_daemon_reload"`
	UnitFilePreset   interface{} `json:"unit_file_preset"`
	UnitFileState    interface{} `json:"unit_file_state"`
}
```

I have chosen some of the fields there. There are much much more (you can check many of them running `systemctl show pacemaker.service` for example. To get the attributes I use the `GetUnitPropertiesContext` from the [dbus API](https://pkg.go.dev/github.com/coreos/go-systemd/v22/dbus).

PD: It can be reviewed. Just waiting first to merge https://github.com/trento-project/agent/pull/277
PD2: @abravosuse I selected the fields above subjectevely, could we ask if they need any other? The whole list can be obtained running `systemctl show corosync.service`


CDimonaco (approved):
lgtm good job

internal/factsengine/gatherers/systemd_v2.go
```diff
@@ -0,0 +1,131 @@
...
+		Message: "error getting systemd unit properties",
+	}
+
+	SystemDDecodingError = entities.FactGatheringError{
+		Type:    "systemd-decoding-error",
+		Message: "error decoding systemd unit status",
+	}
+)
+
+type SystemDGathererV2 struct {
```
rtorrero, line 26: A question about this: should we always define a new struct even when it's the same as the previous version? maybe a type alias makes more sense in such case?

rtorrero (approved):
LGTM, just a small question :+1:

internal/factsengine/gatherers/systemd_v2.go
arbulu89: A new struct is needed. We need to implement the `Gather` function for that.
We could define the new struct as `type SystemDGathererV2 SystemDGatherer`, but I prefer to keep it explicit, to understand the code easier.
I prefer some little duplication in this case

PR #282, opened 2023-10-19
Sapservices gatherer 

CDimonaco:
Gatherer for sapservices file, by default `/usr/sap/sapservices`.

Return each entry of sapservices file as a map, with the kind of startup system used, `systemd` or `sapservices`

### DIFF
--- a/internal/factsengine/gatherers/gatherer.go
+++ b/internal/factsengine/gatherers/gatherer.go
@@ -49,6 +49,9 @@ func StandardGatherers() FactGatherersTree {
 		SapProfilesGathererName: map[string]FactGatherer{
 			"v1": NewDefaultSapProfilesGatherer(),
 		},
+		SapServicesGathererName: map[string]FactGatherer{
+			"v1": NewDefaultSapServicesGatherer(),
+		},
 		SaptuneGathererName: map[string]FactGatherer{
 			"v1": NewDefaultSaptuneGatherer(),
 		},
--- a/internal/factsengine/gatherers/sapservices.go
+++ b/internal/factsengine/gatherers/sapservices.go
@@ -1,0 +1,180 @@
+package gatherers
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"regexp"
+	"strings"
+
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
+	"github.com/trento-project/agent/pkg/factsengine/entities"
+)
+
+type SapServicesStartupKind string
+
+const (
+	SapServicesSystemdStartup  SapServicesStartupKind = "systemctl"
+	SapServicesSapstartStartup SapServicesStartupKind = "sapstartsrv"
+	sapServicesDefaultPath                            = "/usr/sap/sapservices"
+	SapServicesGathererName                           = "sapservices"
+)
+
+// nolint:gochecknoglobals
+var (
+	SapServicesParsingError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error parsing the sap services file",
+	}
+	SapServicesFileError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error reading the sap services file",
+	}
+	SapstartSIDExtractionPattern = regexp.MustCompile(`(?s)pf=([^[:space:]]+)/(.*?)_`)
+	SystemdSIDExtractionPattern  = regexp.MustCompile(`(?s)start SAP(.*?)_.* `)
+)
+
+type SapServicesEntry struct {
+	SID     string                 `json:"sid"`
+	Kind    SapServicesStartupKind `json:"kind"`
+	Content string                 `json:"content"`
+}
+
+func systemdStartup(sapServicesContent string) bool {
+	return strings.Contains(sapServicesContent, "systemctl")
+}
+
+func sapstartStartup(sapServicesContent string) bool {
+	return strings.Contains(sapServicesContent, "sapstartsrv")
+}
+
+func extractSIDFromSystemdService(sapServicesContent string) string {
+	matches := SystemdSIDExtractionPattern.FindStringSubmatch(sapServicesContent)
+	if len(matches) != 2 {
+		return ""
+	}
+	return matches[1]
+}
+
+func extractSIDFromSapstartService(sapServicesContent string) string {
+	matches := SapstartSIDExtractionPattern.FindStringSubmatch(sapServicesContent)
+	if len(matches) != 2 {
+		return ""
+	}
+	return matches[1]
+}
+
+type SapServices struct {
+	fs               afero.Fs
+	servicesFilePath string
+}
+
+func NewSapServicesGatherer(servicesFilePath string, fs afero.Fs) *SapServices {
+	return &SapServices{
+		servicesFilePath: servicesFilePath,
+		fs:               fs,
+	}
+}
+
+func NewDefaultSapServicesGatherer() *SapServices {
+	return &SapServices{servicesFilePath: sapServicesDefaultPath, fs: afero.NewOsFs()}
+}
+
+func (s *SapServices) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting %s facts gathering process", SapServicesGathererName)
+
+	entries, err := s.getSapServicesFileEntries()
+	if err != nil {
+		return nil, err
+	}
+
+	factValues, err := convertSapServicesEntriesToFactValue(entries)
+	if err != nil {
+		return nil, SapServicesParsingError.Wrap(err.Error())
+	}
+
+	for _, requestedFact := range factsRequests {
+		fact := entities.NewFactGatheredWithRequest(requestedFact, factValues)
+
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested %s facts gathered", SapServicesGathererName)
+	return facts, nil
+}
+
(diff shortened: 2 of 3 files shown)


internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,183 @@
...
+var (
+	SapServicesParsingError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error parsing the sap services file",
+	}
+	SapServicesFileError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error reading the sap services file",
+	}
+	SapstartSIDExtractionPattern = regexp.MustCompile(`(?s)pf=([^[:space:]]+)/(.*?)_`)
```
arbulu89, line 34: Nothing against this pattern if it works, but there are many ways of getting the SID.
For example, you could get like: `.*pf=/usr/sap/(.*)/.*`. At the end, the sid is the 3rd folder of the `/usr/sap/SID` path

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,180 @@
...
+func extractSIDFromSapstartService(sapServicesContent string) string {
+	matches := SapstartSIDExtractionPattern.FindStringSubmatch(sapServicesContent)
+	if len(matches) != 2 {
+		return ""
+	}
+	return matches[1]
+}
+
+type SapServices struct {
+	fs               afero.Fs
```
arbulu89, line 69: Do you need afero if you pass a fixed file path?
For example, in the `groups` gatherer you implemented without it

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, err
+	}
+
+	var unmarshalled []interface{}
+	err = json.Unmarshal(marshalled, &unmarshalled)
+	if err != nil {
+		return nil, err
+	}
+
+	return entities.NewFactValue(unmarshalled, entities.WithStringConversion())
```
arbulu89, line 179: I don't think we need `WithStringConversion`, after all, all the values are strings hehe

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,180 @@
...
+
+	for fileScanner.Scan() {
+		scannedLine := fileScanner.Text()
+		if strings.HasPrefix(scannedLine, "#") || scannedLine == "" {
+			continue
+		}
+		entry := SapServicesEntry{}
+		entry.Content = scannedLine
+
+		if systemdStartup(scannedLine) {
```
arbulu89, line 134: I would most probably implement a `getSystemType` function which returns the `Kind` and based on that, do the parsing, maybe even having a map with the extraction functions like `map[string]func(string) string`, so we reduce a bit the repeated code. But I see that this code is more explicit and easy to read

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,180 @@
...
+
+	var entries []SapServicesEntry
+
+	for fileScanner.Scan() {
+		scannedLine := fileScanner.Text()
+		if strings.HasPrefix(scannedLine, "#") || scannedLine == "" {
+			continue
+		}
+		entry := SapServicesEntry{}
+		entry.Content = scannedLine
```
arbulu89, line 132: Would it make sense to story the `content`, `kind` and `sid` in variables and compose the `SapServicesEntry` at the end? So it is visually easier to see the composition.

arbulu89 (commented):
The code logic looks totally correct.
I'm wondering if we should do some integrity check on the line contents. I mean, to check if the line with systemctl or sapstartsrv follows the correct structure

internal/factsengine/gatherers/sapservices.go
CDimonaco: you're right

internal/factsengine/gatherers/sapservices.go
CDimonaco: I was thinking about that, but I prefer some small repetition, the code is really verbose and can help with readability

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,187 @@
...
+var (
+	SapServicesParsingError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error parsing the sap services file",
+	}
+	SapServicesFileError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error reading the sap services file",
+	}
+	SapstartSIDExtractionPattern = regexp.MustCompile(`(?s)pf=([^[:space:]]+)/(.*?)_.*_.*`)
```
arbulu89, line 34: Do you group the `space` thing for any reason? I don't think it is used
By the way, wouldn't be the usage of `.+` more appropriate?
I mean `.*` would match even with empty string

arbulu89 (approved):
Remember to add the gatherer to the list of available gatherers
Maybe you can "mix" the 2 success tests to avoid the "duplicated" warning message

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+// nolint:gochecknoglobals
+var (
+	SapServicesParsingError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error parsing the sap services file",
+	}
+	SapServicesFileError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
```
rtorrero, line 31: I think you forgot to update this with something descriptive of the error

internal/factsengine/gatherers/sapservices_test.go
```diff
@@ -0,0 +1,177 @@
...
+			Gatherer: "sapservices",
+		},
+	}
+
+	result, err := g.Gather(fr)
+	s.Nil(result)
+	s.EqualError(err, "fact gathering error: sap-services-parsing-error - error reading the sap services file: open /usr/sap/sapservices: file does not exist")
+}
+
+func (s *SapServicesGathererSuite) TestGatheringSIDNotIdentifiedSystemd() {
```
rtorrero, line 37: The names of the test functions are IMHO not very clear / descriptive

rtorrero (changes requested):
Thanks @CDimonaco, looks good to me overall but the names of the tests and content of the errors need some fixing

internal/factsengine/gatherers/sapservices.go
CDimonaco: I made the regexp on regex101 as far as I understood it's not a capture but just a more explicit way to express the things, I don't find also the group in the golang extraction

rtorrero (approved):
Approving as we had a chat and its now clear what I meant, feel free to fix and merge when ready :+1:

internal/factsengine/gatherers/gatherer.go
```diff
@@ -68,5 +68,8 @@ func StandardGatherers() FactGatherersTree {
 		VerifyPasswordGathererName: map[string]FactGatherer{
 			"v1": NewDefaultPasswordGatherer(),
 		},
+		SapServicesGathererName: map[string]FactGatherer{
```
arbulu89, line 71: Alphabetical order please

internal/factsengine/gatherers/sapservices.go
arbulu89: Well, in `extractSIDFromSapstartService` you look for a 3 elements size match (which includes, the whole text, 2 groups)

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,187 @@
...
+
+		var kind SapServicesStartupKind
+		var sid string
+
+		if systemdStartup(scannedLine) {
+			kind = SapServicesSystemdStartup
+			extractedSID := extractSIDFromSystemdService(scannedLine)
+			if extractedSID == "" {
+				return nil, SapServicesParsingError.Wrap(
+					fmt.Sprintf("could not extract sid from systemd sap services entry: %s", scannedLine),
```
arbulu89, line 140: I would put `sap` as `SAP` in the text. The same goes for the next error text

internal/factsengine/gatherers/sapservices.go
```diff
@@ -0,0 +1,187 @@
...
+	SapServicesSapstartStartup SapServicesStartupKind = "sapstartsrv"
+	sapServicesDefaultPath                            = "/usr/sap/sapservices"
+	SapServicesGathererName                           = "sapservices"
+)
+
+// nolint:gochecknoglobals
+var (
+	SapServicesParsingError = entities.FactGatheringError{
+		Type:    "sap-services-parsing-error",
+		Message: "error parsing the sap services file",
```
arbulu89, line 28: I would join the `sap services` to `sapservices`. In the `Type` and in the `Message`.
The same for the next errors

arbulu89 (approved):
Looks good! Some cosmetic suggestions.
At least fix the alphabetical ordering of the `getherer.go` file

internal/factsengine/gatherers/sapservices.go
CDimonaco: done

internal/factsengine/gatherers/sapservices.go
CDimonaco: done

internal/factsengine/gatherers/gatherer.go
CDimonaco: done

PR #283, opened 2023-10-25
Add os-release gatherer

rtorrero:
This PR adds a new gatherer for the /etc/os-release file. It will take all the values present in the file and export them as a map.


internal/factsengine/gatherers/osrelease.go
```diff
@@ -0,0 +1,83 @@
...
+func (g *OSReleaseGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
+	facts := []entities.Fact{}
+	log.Infof("Starting %s facts gathering process", OSReleaseGathererName)
+
+	file, err := os.Open(g.osReleaseFilePath)
+	if err != nil {
+		log.Error(err)
+		return facts, OSReleaseFileError.Wrap(err.Error())
+	}
+	defer file.Close()
```
arbulu89, line 52: Let's puts some logs in this `defer`. Just in case. 
As we do in: https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/groups.go#L61

internal/factsengine/gatherers/osrelease.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, requestedFact := range factsRequests {
+		fact := entities.NewFactGatheredWithRequest(requestedFact, osReleaseFactValue)
+		facts = append(facts, fact)
+	}
+
+	log.Infof("Requested %s facts gathered", OSReleaseGathererName)
+	return facts, nil
+}
+
+func mapToFactValue(inputMap map[string]string) entities.FactValue {
```
arbulu89, line 75: Rename this to `mapOSReleaseToFactValue` or similar.
Keep in mind that this function is shared across all the package, and it looks too generic

test/fixtures/gatherers/os-release.basic
```diff
@@ -0,0 +1,12 @@
+NAME="Ubuntu"
+VERSION="20.04.3 LTS (Focal Fossa)"
+ID=ubuntu
```
arbulu89, line 3: I have always seen values wrapped in quotes. Does ubuntu have them without?
It would be nicer to put opensuse :see_no_evil:

arbulu89 (approved):
Looks good!
Add the gatherer to the list of gatherers please.
No need for a new review

test/fixtures/gatherers/os-release.basic
rtorrero: The quotes are not mandatory AFAIK, see some examples: https://manpages.ubuntu.com/manpages/focal/man5/os-release.5.html

I think they are only required when they include values with spaces in them.

I've added some openSUSE love in the mocked data :-)

EMaksy (approved):
Hey man looks good to me! Nice work :)

PR #284, opened 2023-10-26
Mount info gatherer

arbulu89:
Add `mount_info` gatherer. This gatherer returns if a given path is mounted, and its information. Besides that, in case of mounted, it returns the `blkid` `UUID` value of that block.

The `path` to check is the required argument.

If it is not mounted, it returns all the fields as empty string. I needed to do so, because checking if a path (`/usr/sap`) is not mounted (or it is directly in the `root` folder` is the main reason this gatherer exists.

If the path is mounted in nfs-kind of sharing system, the `block_uuid` comes empty

Some outputs:
```
// trento-agent facts gather --gatherer mount_info --argument /usr/sap
#{
  "block_uuid": "f45cf408-efgh-abcd-88ec-2f9269a12f07",
  "fs_type": "xfs",
  "mount_point": "/usr/sap",
  "options": "rw,relatime",
  "source": "/dev/mapper/vg_hana_usrsap-lv_hana_usrsap_0"
} 

// trento-agent facts gather --gatherer mount_info --argument /sapmnt
#{
  "block_uuid": "",
  "fs_type": "cifs",
  "mount_point": "/sapmnt",
  "options": "rw,relatime",
  "source": "10.1.1.10://sapmnt"
} 

```


internal/factsengine/gatherers/mountinfo_test.go
```diff
@@ -0,0 +1,212 @@
...
+		{
+			Name:     "not_mounted",
+			Gatherer: "mount_info",
+			CheckID:  "check1",
+			Argument: "/usr/sap",
+		},
+		{
+			Name:     "shared",
+			Gatherer: "mount_info",
+			CheckID:  "check1",
```
rtorrero, line 85: I've noticed that sometimes we use the same checkid for factrequests in the tests, while other times we put incrementally increased ones and sometimes we don't put it at all. Should we apply some logic when doing so? if on the contrary, it doesn't matter at all for the tests, should we better just remove it?

No need to change anything, just wondering about this.

internal/factsengine/gatherers/mountinfo_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+	expectedResults := []entities.Fact{
+		{
+			Name:    "not_mounted",
+			CheckID: "check1",
+			Value: &entities.FactValueMap{
+				Value: map[string]entities.FactValue{
+					"block_uuid":  &entities.FactValueString{Value: ""},
+					"fs_type":     &entities.FactValueString{Value: ""},
+					"mount_point": &entities.FactValueString{Value: ""},
```
rtorrero, line 114: For now, this is perfect IMHO, in the future if we want to also detect more complex scenarios we could decompose the mountpoint when it's not found and test backwards if the parent exists, e.g. `/usr/sap` passed as argument would first need to check `/usr/sap` and if doesn't find it, it would check `/usr` and if not found, it could then fallback to `/` and just return the details stating in `mount_point` the parent mountpoint of that path.

rtorrero (approved):
Perfect @arbulu89, great work! left some thoughts in the comments but no need to do anything about them

internal/factsengine/gatherers/mountinfo_test.go
arbulu89: It is true that it doesn't add a lot of real value. This is already tested in other places. At most, if you don't use the `NewFactGatheredWithRequest`, it could fail, as you forgot to pass the check id.
For me it is a good practice, as the `CheckID` is a mandatory field for our contract data

internal/factsengine/gatherers/mountinfo_test.go
arbulu89: Let's see if we have such a requirement hehe

EMaksy (commented):
Hey man codewise lgtm :)

PR #287, opened 2023-11-02
Add testpackage to golangci

arbulu89:
A good reminder that we need to follow the `package_test` rule!
The CI should pass once we have fixed all the packages in other PRs.
If for some reason we need to exclude some package, we can add exceptions, but by general rule, all of them should be compliant.

https://github.com/maratori/testpackage

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -45,6 +45,7 @@ linters:
     - unconvert
     - unparam
     - wastedassign
+    - testpackage
 
 issues:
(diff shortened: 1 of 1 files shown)


EMaksy (commented):
Good idea :)

CDimonaco (approved):
Absolutely a good idea, lgtm

rtorrero (approved):
Good call 👍

PR #288, opened 2023-11-02
Refactor collector package tests

arbulu89:
Refactor collector package tests

### DIFF
--- a/internal/discovery/collector/client_test.go
+++ b/internal/discovery/collector/client_test.go
@@ -18,23 +18,30 @@ const (
 
 type CollectorClientTestSuite struct {
 	suite.Suite
+	collectorClient *collector.Collector
+	httpClient      *http.Client
 }
 
 func TestCollectorClientTestSuite(t *testing.T) {
 	suite.Run(t, new(CollectorClientTestSuite))
 }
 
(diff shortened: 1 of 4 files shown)


internal/discovery/collector/client_test.go
```diff
@@ -24,12 +25,15 @@ func TestCollectorClientTestSuite(t *testing.T) {
 }
 
 func (suite *CollectorClientTestSuite) TestCollectorClientPublishingSuccess() {
-	collectorClient := NewCollectorClient(
-		&Config{
+	httpClient := http.DefaultClient
```
nelsonkopliku, line 28: Not a biggie, however what about having `httpClient` set in the suite and pass it as the second argument of `NewCollectorClient`?
```go
collectorClient := collector.NewCollectorClient(
    &collector.Config{/** ... */},
    suite.httpClient,
)
```
as in publishing_test.go?

arbulu89: No problem! I can change that

arbulu89: Done

nelsonkopliku (approved):

EMaksy (commented):
Looks good :)

CDimonaco (approved):

PR #291, opened 2023-11-13
Refactor cloud package tests

arbulu89:
Refactor `cloud` package tests

### DIFF
--- a/internal/core/cloud/metadata.go
+++ b/internal/core/cloud/metadata.go
@@ -181,7 +181,7 @@ func (i *Identifier) IdentifyCloudProvider() (string, error) {
 	return "", nil
 }
 
-func NewCloudInstance(commandExecutor utils.CommandExecutor) (*Instance, error) {
+func NewCloudInstance(commandExecutor utils.CommandExecutor, client HTTPClient) (*Instance, error) {
 	var err error
 	var cloudMetadata interface{}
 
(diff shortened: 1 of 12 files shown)


internal/core/cloud/metadata_test.go
```diff
@@ -393,7 +390,7 @@ func (suite *CloudMetadataTestSuite) TestNewInstanceNutanix() {
 		dmidecodeNutanix(), nil,
 	)
 
-	c, err := NewCloudInstance(mockCommand)
+	c, err := cloud.NewCloudInstance(mockCommand, nil)
```
nelsonkopliku, line 393: I have mixed feelings about passing a `nil` client in these tests where we know client is not needed. The `NewCloudInstance` API looks a bit inconsistent and leaking a bit the internals.

What about passing an instance of a client and expecting that no requests on that client would be executed?

arbulu89: Okey!

arbulu89: Done

nelsonkopliku (approved):

EMaksy (commented):
Hey could not find anything :)

internal/core/cloud/metadata.go
```diff
@@ -199,18 +199,18 @@ func NewCloudInstance(commandExecutor utils.CommandExecutor) (*Instance, error)
 
 	switch provider {
 	case Azure:
-		cloudMetadata, err = NewAzureMetadata()
+		cloudMetadata, err = NewAzureMetadata(client)
```
CDimonaco, line 202: Could we add { } to the cases? just to increase readability

CDimonaco (approved):
lgtm, just the comment, my ocd is triggered 👀

PR #292, opened 2023-11-13
Refactor cluster package tests

arbulu89:
Refactor cluster package tests.


internal/core/cluster/sbd_test.go
```diff
@@ -55,98 +56,19 @@ sbd failed; please check the logs.`
 	return []byte(output)
 }
 
-func (suite *SbdTestSuite) TestSbdDump() {
```
arbulu89, line 58: Removing this tests as this part of the code is already tested in other public function tests

internal/core/cluster/sbd.go
```diff
@@ -54,9 +53,8 @@ type SBDNode struct {
 	Status string
 }
 
-func NewSBD(executor utils.CommandExecutor, cluster, sbdPath, sbdConfigPath string) (SBD, error) {
+func NewSBD(executor utils.CommandExecutor, sbdPath, sbdConfigPath string) (SBD, error) {
 	var s = SBD{
-		cluster: cluster,
```
arbulu89, line 59: It is not used at all.
I could keep it and exclude it's comparison in the assert, but I think removing it is better.

nelsonkopliku (approved):
:eyes:

EMaksy (commented):
LGTM :)

rtorrero (approved):
LGTM

CDimonaco (approved):

PR #300, opened 2023-11-21
Http client refactor

CDimonaco:
Http Client refactor

- Removed usage of http.DefaultClient in production code, allowed only in testing
- Removed usage of http.NewRequest, using http.NewRequestWithContext
- Context injection in the discovery interface
- Context injection in soap/http clients
- Context propagation


nelsonkopliku (approved):
:eyes:

internal/core/cloud/azure.go
```diff
@@ -220,7 +221,11 @@ func NewAzureMetadata(client HTTPClient) (*AzureMetadata, error) {
 		},
 	}
 
-	req, _ := http.NewRequest(http.MethodGet, fmt.Sprintf("http://%s/metadata/instance", azureAPIAddress), nil)
+	req, _ := http.NewRequestWithContext(
+		ctx, http.MethodGet,
```
arbulu89, line 225: I would move `http.MethodGet` to a new line

internal/core/cloud/aws_test.go
```diff
@@ -24,6 +25,7 @@ func TestAWSMetadataTestSuite(t *testing.T) {
 }
 
 func (suite *AWSMetadataTestSuite) TestNewAWSMetadata() {
+	ctx := context.TODO()
```
arbulu89, line 28: Is this `TODO` the preferred practice for tests?

internal/discovery/cloud.go
```diff
@@ -36,14 +37,14 @@ func (d CloudDiscovery) GetInterval() time.Duration {
 	return d.interval
 }
 
-func (d CloudDiscovery) Discover() (string, error) {
-	client := &http.Client{Transport: &http.Transport{Proxy: nil}}
-	cloudData, err := cloud.NewCloudInstance(utils.Executor{}, client)
+func (d CloudDiscovery) Discover(ctx context.Context) (string, error) {
```
arbulu89, line 40: Just a comment, don't change it XD
It would've been nice to have something like `Discover` and `DiscoverWithContext`, `Publish` and `PublishWithContext`, it looks much more golang like hehe

arbulu89 (approved):
Great!
Thank you so much!

internal/discovery/cloud.go
CDimonaco: I was thinking about that, but we need two interfaces, or two methods on the same interface, in the end the correct way is with context, at least in this case, is a loop started in a goroutine that already has context (in the agent file)

internal/core/cloud/aws_test.go
CDimonaco: Yep, it's an "empty context" made just to satisfy the interface

You could also use test context but it's not the preferred way, it's like mixing things

internal/core/cloud/azure.go
CDimonaco: done

PR #305, opened 2023-11-29
Fix verify_password gatherer for scenarios where there is not hash

arbulu89:
https://bugzilla.suse.com/show_bug.cgi?id=1217543

```
>  panic: runtime error: index out of range [2] with length 1
> 
> goroutine 264 [running]:
>  github.com/trento-project/agent/internal/factsengine/gatherers.(VerifyPasswordGatherer).getSalt(0xc000484030, {0xc0005ab5a0, 0x9})
>  github.com/trento-project/agent/internal/factsengine/gatherers/verifypassword.go:111 +0x21c
>  github.com/trento-project/agent/internal/factsengine/gatherers.(VerifyPasswordGatherer).Gather(0xc000484030?, {0xc00030f100, 0x1, 0x1?})
>  github.com/trento-project/agent/internal/factsengine/gatherers/verifypassword.go:68 +0x1e5
>  github.com/trento-project/agent/internal/factsengine.gatherFacts.func1()
>  github.com/trento-project/agent/internal/factsengine/gathering.go:44 +0xb9
>  golang.org/x/sync/errgroup.(Group).Go.func1()
>  golang.org/x/[sync@v0.2.0](mailto:sync@v0.2.0)/errgroup/errgroup.go:75 +0x64
>  created by golang.org/x/sync/errgroup.(Group).Go
>  golang.org/x/[sync@v0.2.0](mailto:sync@v0.2.0)/errgroup/errgroup.go:72 +0xa5
```

### DIFF
--- a/internal/factsengine/gatherers/verifypassword_test.go
+++ b/internal/factsengine/gatherers/verifypassword_test.go
@@ -88,8 +88,8 @@ func (suite *PasswordTestSuite) TestPasswordGatherNotEqual() {
 	suite.ElementsMatch(expectedResults, factResults)
 }
 
-func (suite *PasswordTestSuite) TestPasswordGatherCryptError() {
-	shadow := []byte("hacluster:$aaaa$aaaa")
+func (suite *PasswordTestSuite) TestPasswordGatherNotEncrypted() {
+	shadow := []byte("hacluster:!:19029::::::")
 
 	suite.mockExecutor.On("Exec", "getent", "shadow", "hacluster").Return(
 		shadow, nil)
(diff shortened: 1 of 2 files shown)


internal/factsengine/gatherers/verifypassword_test.go
```diff
@@ -88,8 +88,8 @@ func (suite *PasswordTestSuite) TestPasswordGatherNotEqual() {
 	suite.ElementsMatch(expectedResults, factResults)
 }
 
-func (suite *PasswordTestSuite) TestPasswordGatherCryptError() {
-	shadow := []byte("hacluster:$aaaa$aaaa")
+func (suite *PasswordTestSuite) TestPasswordGatherNotEncrypted() {
+	shadow := []byte("hacluster:!:19029::::::")
```
rtorrero, line 92: Hey @arbulu89, just wondering about this: why is the test name "...NotEncrypted"? seems to me it's more simply about the case where no password is set (rather than not encrypted).

Also, about the exclamation mark, I think an exclamation mark at the start of the field means simply "account disabled" while an exclamation mark in the content of the hash is an invalid hash (as it is an invalid character in this context)

arbulu89: Yes, we could improve the naming of the text.
About what the `!` means, based on this:

rtorrero: I think I understand my doubts now, there is a "common practice" (even documented in `man 5 shadow`) that says that if you add a `!` as the first character of a password it means "password disabled" but in reality, adding it anywhere will block the password anyways. So effectively, a password containing just `!` or `*` means: "no password, and also no login allowed".

CDimonaco (approved):

dottorblaster (approved):
Sounds good :+1:

PR #313, opened 2024-01-15
Discover FQDN inside the host discovery loop

dottorblaster:
Send an FQDN for the host inside the host discovery loop. If no FQDN is found the field is omitted from the payload.

### DIFF
--- a/internal/discovery/host.go
+++ b/internal/discovery/host.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/Showmax/go-fqdn"
 	"github.com/shirou/gopsutil/cpu"
 	"github.com/shirou/gopsutil/host"
 	"github.com/shirou/gopsutil/mem"
@@ -26,7 +27,11 @@ type HostDiscovery struct {
 	interval        time.Duration
 }
 
-func NewHostDiscovery(collectorClient collector.Client, hostname string, config DiscoveriesConfig) Discovery {
+func NewHostDiscovery(
+	collectorClient collector.Client,
+	hostname string,
+	config DiscoveriesConfig,
+) Discovery {
 	return HostDiscovery{
 		id:              HostDiscoveryID,
 		collectorClient: collectorClient,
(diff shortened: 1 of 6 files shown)


internal/discovery/mocks/discovered_host_mock.go
```diff
@@ -3,14 +3,17 @@ package mocks
...
-		InstallationSource: "Community",
+		OSVersion:                "15-SP2",
+		HostIPAddresses:          []string{"10.1.1.4", "10.1.1.5", "10.1.1.6"},
+		HostName:                 "thehostnamewherethediscoveryhappened",
+		CPUCount:                 2,
+		SocketCount:              1,
+		TotalMemoryMB:            4096,
+		AgentVersion:             "trento-agent-version",
+		InstallationSource:       "Community",
+		FullyQualifiedDomainName: &fqdn,
```
nelsonkopliku, line 17: What about just using the plain string here? :eyes:

internal/agent/agent.go
```diff
@@ -31,11 +31,12 @@ type Agent struct {
 }
 
 type Config struct {
-	AgentID           string
-	InstanceName      string
-	DiscoveriesConfig *discovery.DiscoveriesConfig
-	FactsServiceURL   string
-	PluginsFolder     string
+	AgentID                  string
+	FullyQualifiedDomainName string
```
nelsonkopliku, line 35: No need for change, just to say I'd be fine also with `FQDN`/`Fqdn`

internal/discovery/host.go
```diff
@@ -50,15 +52,21 @@ func (d HostDiscovery) Discover(ctx context.Context) (string, error) {
 		return "", err
 	}
 
+	var discoveredHostFQDN *string
```
nelsonkopliku, line 55: do we need this extra local variable? `d.fqdn` might come nullish (empty string) anyway from the caller, right? Or am I missing something?

nelsonkopliku (approved):
LGMT.

Just a couple of minor things/doubts.

jamie-suse (approved):

internal/discovery/host.go
```diff
@@ -93,6 +99,20 @@ func getHostIPAddresses() ([]string, error) {
 	return ipAddrList, nil
 }
 
+func getHostFQDN() *string {
```
nelsonkopliku, line 102: Well done! It makes sense actually discovering this in the discovery.
Yes, it did work getting it at config build time, but this sounds more appropriate :+1:

dottorblaster: Yeah in the end I just went straight to the point mainly because the FQDN can actually change and we'd lose that modification if such an information was collected at the start

PR #321, opened 2024-02-14
Send sbd data in diskless scenario

arbulu89:
Update cluster discovery part to allow sending SBD data in diskless scenario.
In this case, even if the cluster doesn't have any SBD resource, the cluster can be running a SBD diskless fencing type.
The code removes that conditions. It will still print an error message is something goes wrong during SBD discovery and the SBD resource is configured. The code won't crash anymore, as not having SBD is not critical always, and the web part handles all this scenarios. Better to send the payload in any case.

I have removed some unused code as well.

The rationales:
SBD diskless consists on having SBD configured and running, but without any SBD device or resource.
If somebody wants to read further: https://documentation.suse.com/sle-ha/15-SP5/html/SLE-HA-all/cha-ha-storage-protect.html#sec-ha-storage-protect-diskless-sbd

### DIFF
--- a/internal/core/cluster/cluster.go
+++ b/internal/core/cluster/cluster.go
@@ -5,7 +5,6 @@ import (
 	"encoding/hex"
 	"io"
 	"os"
-	"strconv"
 	"strings"
 
 	// These packages were originally imported from github.com/ClusterLabs/ha_cluster_exporter/collector/pacemaker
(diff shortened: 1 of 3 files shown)


rtorrero (approved):
LGTM @arbulu89, thanks!

PR #322, opened 2024-02-15
Update saptune discovery interval

rtorrero:
This PR:
 - adjusts the saptune discovery interval from 10s to 900s by default
 - includes an example on how to override the default in the bundled configuration


arbulu89 (approved):
:top:

EMaksy (approved):
Nice!

PR #325, opened 2024-03-04
Facts cache

arbulu89:
Add basic facts cache. It is designed to store facts inside the `Gather` function, is the provided gatherer implements the `WithCache` function (coming from `FactGathererWithCache`).

The cache storage has a locking mechanism to avoid race conditions, as the gathering happens in a parallel.

The usage is pretty simple.
- It needs a cache entry name
- A function to get the value that is stored in the function with a specific function signature
- Optional arguments for that function

The memoization is achieved using [singleflight](https://pkg.go.dev/golang.org/x/sync/singleflight#Group.Do)
Basically based on https://github.com/kofalt/go-memoize/blob/master/memoize.go, but we need some adjustments so I have created a custom version (and we don't import many other packages only for this as well)

### DIFF
--- a/internal/factsengine/factscache/cache.go
+++ b/internal/factsengine/factscache/cache.go
@@ -1,0 +1,68 @@
+package factscache
+
+import (
+	"sync"
+
+	log "github.com/sirupsen/logrus"
+)
+
+type FactsCache struct {
+	entries map[string]Entry
+	lock    sync.Mutex
+}
+
+type Entry struct {
+	content interface{}
+	err     error
+}
+
+func NewFactsCache() *FactsCache {
+	return &FactsCache{
+		entries: make(map[string]Entry),
+		lock:    sync.Mutex{},
+	}
+}
+
+// Entries returns the cached entries list
+func (c *FactsCache) Entries() []string {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	keys := []string{}
+	for key := range c.entries {
+		keys = append(keys, key)
+	}
+	return keys
+}
+
+// GetOrUpdate returns the cached result providing an entry name
+// or runs the updateFunc to generate the entry.
+// It locks its usage, so only one user at a time uses it
+func (c *FactsCache) GetOrUpdate(
+	entry string,
+	udpateFunc func(args ...interface{}) (interface{}, error),
+	updateFuncArgs ...interface{},
+) (interface{}, error) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	cacheEntry, hit := c.entries[entry]
+	if hit {
+		log.Debugf("Value for entry %s already cached", entry)
+		return cacheEntry.content, cacheEntry.err
+	}
+
+	content, err := udpateFunc(updateFuncArgs...)
+	c.entries[entry] = Entry{
+		content: content,
+		err:     err,
+	}
+
+	if err != nil {
+		log.Debugf("New value with error set for entry %s", entry)
+		return content, err
+	}
+
+	log.Debugf("New value for entry %s set", entry)
+	return content, err
+}
--- a/internal/factsengine/factscache/cache.go
+++ b/internal/factsengine/factscache/cache.go
@@ -3,12 +3,16 @@ package factscache
 import (
 	"sync"
 
+	"golang.org/x/sync/singleflight"
+
 	log "github.com/sirupsen/logrus"
 )
 
+type UpdateCacheFunc func(args ...interface{}) (interface{}, error)
+
 type FactsCache struct {
-	entries map[string]Entry
-	lock    sync.Mutex
+	entries sync.Map
+	group   singleflight.Group
 }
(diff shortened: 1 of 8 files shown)


internal/factsengine/gatherers/cibadmin.go
```diff
@@ -25,26 +27,51 @@ var (
...
+func (g *CibAdminGatherer) SetCache(cache *factscache.FactsCache) {
+	g.cache = cache
+}
+
 func (g *CibAdminGatherer) Gather(factsRequests []entities.FactRequest) ([]entities.Fact, error) {
 	log.Infof("Starting %s facts gathering process", CibAdminGathererName)
+	var content interface{}
+	var err error
+
+	updateCacheFunc := func(args ...interface{}) (interface{}, error) {
```
arbulu89, line 53: We could make the cache usage mandatory as well

internal/factsengine/factscache/cache.go
```diff
@@ -0,0 +1,68 @@
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	cacheEntry, hit := c.entries[entry]
+	if hit {
+		log.Debugf("Value for entry %s already cached", entry)
+		return cacheEntry.content, cacheEntry.err
+	}
+
+	content, err := udpateFunc(updateFuncArgs...)
```
arbulu89, line 55: We need to pass the `updateFunc` function, as we want to make the whole cache usage of one shot to use the lock properly.
Passing the value is not enough, as getting the value might be delayed.

internal/factsengine/factscache/cache_test.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	value, err := cache.GetOrUpdate("entry", updateFunc, 1, "text")
+
+	suite.Equal("1_text", value)
+	suite.NoError(err)
+}
+
+// nolint:errcheck
+func (suite *FactsCacheTestSuite) TestGetOrUpdateCacheConcurrent() {
```
arbulu89, line 105: I'm not entirely satisfied with this concurrency test, but I found it really needed, and it only needs 100ms.
Let me know if there is any other better way

internal/factsengine/factscache/cache_test.go
```diff
@@ -0,0 +1,173 @@
...
+	cache := factscache.NewFactsCache()
+	g := errgroup.Group{}
+
+	updateFunc := func(args ...interface{}) (interface{}, error) {
+		value, _ := args[0].(string)
```
rtorrero, line 110: Again, maybe it has little value but it could make it a bit more clear if the updateFunc uses an int and increments the value each time there is a cache miss.

internal/factsengine/factscache/cache_test.go
```diff
@@ -0,0 +1,173 @@
...
+		return nil
+	})
+	g.Go(func() error {
+		value, _ := cache.GetOrUpdate("entry2", updateFunc, "value3")
+		castedValue, _ := value.(string)
+		suite.Equal("value3", castedValue)
+		return nil
+	})
+	time.Sleep(50 * time.Millisecond)
+	// The next 2 calls return the memoized value
```
rtorrero, line 128: Maybe it's a bit overkill, but just wondering if we should we test that these 2 following calls are not blocking for 100ms and instead return immediately?

rtorrero (commented):
Good stuff @arbulu89. I like the approach, only thing I'm wondering is if we should care for cases like `sapcontrol` where we would benefit from some caching that is both useful for discoveries and fact gathering: e.g. a fact gathering might force to re-request something that was just requested for a discovery mechanism.

internal/factsengine/factscache/cache.go
```diff
@@ -0,0 +1,99 @@
...
+		// nolint:forcetypeassert
+		keys = append(keys, key.(string))
+		return true
+	})
+	return keys
+}
+
+// GetOrUpdate returns the cached result providing an entry name
+// or runs the updateFunc to generate the entry.
+// It locks its usage, so only one user at a time uses it
```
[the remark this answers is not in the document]
arbulu89: It only blocks when the function used with a key that right now is being executed.
I will update the comment

internal/factsengine/factscache/cache_test.go
arbulu89: you cannot do that in the `updateFunc` function. This function doesn't even know that it is being called in this cache method.
It is possible to return if the you are getting a cached value or not, but what real value would that have? we don't do any different in any of the cases

internal/factsengine/factscache/cache_test.go
arbulu89: how can you test that? I mean, the time that the test needs. Is that even possible?
Besides, they don't return immediately, the get blocked until the first one is finished, and both return at the exact same moment with the value from the first execution.

CDimonaco (approved):
LGTM

PR #326, opened 2024-03-05
Improve crash on failed agent initialization

rtorrero:
This PR:
 - Ensures that the agent doesn't start the discovery loop until the subscription to the facts engine is working
 
NOTE: I also explored what would happen on wrong configuration files, e.g. following the example of the slack conversation from TRNT-2814, I removed `facts-service-url` and this seems that was already causing the agent to crash without a heartbeat nor discovery loops starting
 
(Addresses TRNT-2814)

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -74,7 +74,23 @@ func GetAgentID(fileSystem afero.Fs) (string, error) {
 
 // Start the Agent. This will start the discovery ticker and the heartbeat ticker
 func (a *Agent) Start(ctx context.Context) error {
+	gathererRegistry := gatherers.NewRegistry(gatherers.StandardGatherers())
 	g, groupCtx := errgroup.WithContext(ctx)
+
+	c := factsengine.NewFactsEngine(a.config.AgentID, a.config.FactsServiceURL, *gathererRegistry)
+	log.Info("Starting fact gathering service...")
+	if err := c.Subscribe(); err != nil {
+		return err
+	}
+	g.Go(func() error {
+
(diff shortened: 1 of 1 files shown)


internal/agent/agent.go
```diff
@@ -74,8 +74,24 @@ func GetAgentID(fileSystem afero.Fs) (string, error) {
 
 // Start the Agent. This will start the discovery ticker and the heartbeat ticker
 func (a *Agent) Start(ctx context.Context) error {
+	gathererRegistry := gatherers.NewRegistry(gatherers.StandardGatherers())
```
arbulu89, line 77: Maybe we could move the `g, groupCtx` initialization after the  error checking, before ling `85`. As it won't be used if it fails before

internal/agent/agent.go
```diff
@@ -74,8 +74,24 @@ func GetAgentID(fileSystem afero.Fs) (string, error) {
...
 	g, groupCtx := errgroup.WithContext(ctx)
 
+	c := factsengine.NewFactsEngine(a.config.AgentID, a.config.FactsServiceURL, *gathererRegistry)
+	log.Info("Starting fact gathering service...")
+	if err := c.Subscribe(); err != nil {
+		return err
+	}
+	g.Go(func() error {
+
```
arbulu89, line 86: Can we remove this 86 empty line?
And maybe put a new line before `g.Go`

arbulu89 (commented):
@rtorrero The code looks ok.
Can we add a test for the failed scenario?

arbulu89 (approved):

PR #328, opened 2024-03-07
Add cache usage to sapcontrol gatherer

arbulu89:
Add factscache usage to sapcontrol gatherer, as sapcontrol data needs to be used in the future new gatherer to get ASCS/ERS clusters data.

It stores the entry in the cache as `sapcontrol:${method}:${sid}:${instance}`. for example: `sapcontrol:GetProcessList:PRD:00`.

This removes the need for a local cache as well.

### DIFF
--- a/internal/factsengine/gatherers/sapcontrol.go
+++ b/internal/factsengine/gatherers/sapcontrol.go
@@ -11,11 +11,13 @@ import (
 	"github.com/spf13/afero"
 	"github.com/trento-project/agent/internal/core/sapsystem"
 	"github.com/trento-project/agent/internal/core/sapsystem/sapcontrolapi"
+	"github.com/trento-project/agent/internal/factsengine/factscache"
 	"github.com/trento-project/agent/pkg/factsengine/entities"
 )
 
 const (
-	SapControlGathererName = "sapcontrol"
+	SapControlGathererName  = "sapcontrol"
+	SapControlGathererCache = "sapcontrol"
 )
 
 // nolint:gochecknoglobals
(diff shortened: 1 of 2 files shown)


internal/factsengine/gatherers/sapcontrol.go
```diff
@@ -89,24 +91,58 @@ type SapControlInstance struct {
...
 		fs:         fs,
+		cache:      cache,
+	}
+}
+
+func (s *SapControlGatherer) SetCache(cache *factscache.FactsCache) {
+	s.cache = cache
+}
+
+func memoizeSapcontrol(args ...interface{}) (interface{}, error) {
```
arbulu89, line 119: I cannot add a test for this...
In reality, this `!ok` shouldn't never happen as this function is only called internally with known arguments.
But I just added the cases for some unexpected error, that I cannot forsee now

EMaksy (commented):
Hey, hey, no go export, but the logic seems to be fully fine! 
LGTM

CDimonaco (approved):

internal/factsengine/gatherers/sapcontrol.go
```diff
@@ -89,24 +91,58 @@ type SapControlInstance struct {
...
+	ctx, ok := args[0].(context.Context)
+	if !ok {
+		return nil, ImplementationError.Wrap("error using memoizeSapcontrol. Context must be 1st argument")
+	}
+
+	webService, ok := args[1].(sapcontrolapi.WebServiceConnector)
+	if !ok {
+		return nil, ImplementationError.Wrap("error using memoizeSapcontrol. WebServiceConnecter must be 2nd argument")
```
rtorrero, line 127: WebServiceConnectOr (typo)

internal/factsengine/gatherers/sapcontrol.go
```diff
@@ -89,24 +91,58 @@ type SapControlInstance struct {
...
+	}
+
+	instanceNumber, ok := args[2].(string)
+	if !ok {
+		return nil, ImplementationError.Wrap("error using memoizeSapcontrol. string must be 3rd argument")
```
rtorrero, line 132: Capitalize after `.` (also in line 137)

rtorrero (approved):
LGTM

internal/factsengine/gatherers/sapcontrol.go
arbulu89: This is intentional. `string` type is in small case. The others are in Capital case, as they are described this way in the code

PR #329, opened 2024-03-11
Add new ascsers_cluster gatherer

arbulu89:
New `ascsers_cluster` gatherer, which gathers details about this type of clusters. **Open for other name options**
The gatherer doesn't require any argument.

To use, simply:
```yaml
facts:
  - name: cluster_details
    gatherer: ascsers_cluster
```

This is what it returns (once approved, this will be commented in the gatherers docs):

```ts
#{
  "PRD": #{
    "ensa_version": "ensa1/ensa2/unknown",
    "instances": [
      #{
         "name" : "ASCS00", // 2nd value of the InstanceName value
         "instance_number: "00", 
         "virtual_hostname": "sapascs00", // 3rd value of the InstanceName value
         "filesystem_based": true/false, // if the instance resource group has a FileSystem resource
         "local": true/false // if the instance is running locally (sapcontrol returns a positive response for GetProcessList)
      },...
   ]
  },
  "DEV": #{
     "ensa_version": "ensa1/ensa2/unknown",
     "instances": [
...
  }
}
```

FYI: @yeoldegrove


nelsonkopliku (approved):
LGTM, I guess :smile: 

With regards to naming, I often captured `managed_systems` in the discussions about this.

What about something like:
- `ascsers_cluster_managed_systems`
- `ascsers_managed_systems`
- `cluster_managed_systems`

Great job!

EMaksy (approved):
Hey from my side looks good :+1:

arbulu89:
@yeoldegrove thank you so much!
the changes look good to me. I'm waiting for some additional review from my team, but i think we can merge it the next days for sure

nelsonkopliku (approved):
LGTM

internal/factsengine/gatherers/ascsers_cluster.go
```diff
@@ -0,0 +1,246 @@
...
+		cacheEntry,
+		memoizeSapcontrol,
+		ctx,
+		webService,
+		instanceNumber,
+		mapGetProcessList,
+	)
+
+	processes, ok := output.([]*sapcontrolapi.OSProcess)
+	if err != nil || !ok {
```
CDimonaco, line 217: This could panic, because in the second or condition err if nil and it's used in logger

internal/factsengine/gatherers/ascsers_cluster.go
```diff
@@ -0,0 +1,246 @@
...
+		log.Warnf("error requesting GetProcessList information: %s", err.Error())
+		return EnsaUnknown, false
+	}
+
+	for _, process := range processes {
+		if process.Name == "enserver" || process.Name == "enrepserver" {
```
CDimonaco, line 223: Could we add a fallthrough switch for this checks?

CDimonaco (commented):
Good job man, I left some comments

internal/factsengine/gatherers/ascsers_cluster.go
arbulu89: How would you use it?
I don't see where I could use the `fallthrough`.
I could write it as a switch like this, but it doesn't look much better:
```

switch process.Name {
case "enserver":
	return Ensa1, true
case "enrepserver":
	return Ensa1, true
case "enq_server":
	return Ensa2, true
case "enq_replicator":
	return Ensa2, true
}
		}
```

CDimonaco: Yeah Sorry, I used the wrong words, I mean we could use multiple values for the case statement,  it's evaluated like an or condition

CDimonaco (approved):

internal/factsengine/gatherers/ascsers_cluster.go
```diff
@@ -0,0 +1,246 @@
...
+						return nil, fmt.Errorf("incorrect InstanceName property value: %s", instanceAttribute.Value)
+					}
+
+					instanceGroupFound = true
+					resourceGroup = group.ID
+					resourceInstance = primitive.ID
+					sid = values[0]
+					instanceName = values[1]
+					virtualHostname = values[2]
+					instanceNumber = instanceName[len(instanceName)-2:]
```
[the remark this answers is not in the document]
arbulu89: Done

internal/factsengine/gatherers/ascsers_cluster.go
arbulu89: Done

internal/factsengine/gatherers/ascsers_cluster.go
arbulu89: Done

PR #333, opened 2024-06-05
Updated osutil to fix missing dependency

CDimonaco:
The package https://github.com/tredoe/fileutil is no longer available, this was a dependency of https://github.com/tredoe/osutil/.

We use `osutil` package and we need to update to the latest version, where they fixed the dependency resolution, without this update, compilation and building will fail.


dottorblaster (approved):
LGTM

PR #334, opened 2024-06-06
Fix install-agent.sh with --use-tgz

### LINKED ISSUE #222: wrong path in install-agent.sh
using the parameter --use-tgz failes because of an wrong path in

https://github.com/trento-project/agent/blob/main/install-agent.sh#L200

it should be:

`mv trento-agent.service ${sysd_dir}/trento-agent.service`

stefanotorresi:
fixes #222

tested manually, since no automated tests are present for the install script.

in draft due to merge freeze currently in place.


arbulu89 (approved):
Thank you!

Issue #221, opened 2023-04-27
document the defaut mqs location


GER-Space:
Update the documentation to the right facts server URL (userame, password, location). guest:guest@localhost is wrong on all levels and confusing to anyone setting it up in a non-dev environment. 

 trento:trento@\<trento-server\> would be a better lead.

PR #335, opened 2024-06-06
Better document the config example

stefanotorresi:
as suggested in #221

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,10 @@ Note that order represents priority
 
 api-key: <api-key-generated-from-the-server>
 server-ip: https://localhost
+facts-service-url: amqp://guest:guest@localhost:5672
 ```
+
+Please refer to the [default configuration file](packaging/config/agent.yaml) for more detailed information on the various settings.
 
 ## Environment Variables
 
--- a/packaging/config/agent.yaml
+++ b/packaging/config/agent.yaml
@@ -86,13 +86,15 @@
 
 ## Configure the api-key for trento agent to use in the communication with Trento control plane
 ## Required.
-## Find the api key at <your-trento-installation-url>/settings 
+## Find the api key at <your-trento-installation-url>/settings
 
 # api-key: <the-api-key-generated-by-trento>
 
 ###############################################################################
 
 ## Facts gathering service URL
+## URL to an AMQP service handling communication between the Trento Agent and Server components
+## For more details on the URI scheme, please refer to https://www.rabbitmq.com/docs/uri-spec
 
 # facts-service-url: amqp://guest:guest@localhost:5672
 


README.md
```diff
@@ -191,8 +191,11 @@ Note that order represents priority
 
 api-key: <api-key-generated-from-the-server>
 server-ip: https://localhost
+facts-service-url: amqp://guest:guest@localhost:5672
```
EMaksy, line 194: I think we need to adjust this part.
The facts-service-URL depends on the configuration of RabbitMQ.

Usually we create a user with a password:
```
rabbitmqctl add_user trento_user trento_user_password
```
and a virtual host:
```
rabbitmqctl add_vhost vhost
```
This params are a must for the facts service url, so we need to adjust the example.

```facts-service-url: amqp://guest:guest@localhost:5672``` -> ```facts-service-url: amqp://guest:guest_password@localhost:5672/vhost```

or we use the same facts-service-url which is mentioned in the manual installation guide.
```
amqp://trento_user:trento_user_password@trento.example.com:5672/vhost
```

File: packaging/config/agent.yaml [hunk 2]
```diff
@@ -93,5 +93,7 @@
 ###############################################################################
 
 ## Facts gathering service URL
+## URL to an AMQP service handling communication between the Trento Agent and Server components
+## For more details on the URI scheme, please refer to https://www.rabbitmq.com/docs/uri-spec
 
-# facts-service-url: amqp://guest:guest@localhost:5672
\ No newline at end of file
+# facts-service-url: amqp://guest:guest@localhost:5672
```
EMaksy, line 99: Also here:
`facts-service-url: amqp://guest:guest@localhost:5672` -->  `facts-service-url: amqp://guest:guest_password@localhost:5672/vhost`

EMaksy: See example [from official docs](https://www.rabbitmq.com/docs/uri-spec#appendix-a-examples):
```
amqp://user:pass@host:10000/vhost
```

EMaksy (changes requested):
Hey @stefanotorresi thanks for changing this!
Everything looks good, but we need to change facts-service-URL, I left some explanations as comments why :+1:

packaging/config/agent.yaml
stefanotorresi: my rationale with leaving as it is that in the default configuration we need to put the default values, as in:
https://github.com/trento-project/agent/blob/58fd08c17c2152137c5eb6d7f602abc8a836ce06/cmd/start.go#L124

and I don't really want to change the current default. I think the explanation and the reference should be enough to fully understand the URL scheme, isn't it?

EMaksy: Oh, I see :)
Yeah, sure :)

EMaksy (approved):

EMaksy (approved):
LGTM!

PR #336, opened 2024-06-20
Add ip_addresses_netmasks field to host discovery

rtorrero:
This PR adds the `ip_addresses_netmasks` field to the hosts discovery. We previously discarded the netmask to prepare the content for the `ip_addresses` field. Now, instead of discarding it, we send it together with the IP address in CIDR notation:
```
...
ip_addresses: ["192.168.2.3"],  //This field hasn't been changed
ip_addresses_netmasks: ["192.168.2.3/24"],
...
```

### DIFF
--- a/internal/core/hosts/discovered_host.go
+++ b/internal/core/hosts/discovered_host.go
@@ -3,6 +3,7 @@ package hosts
 type DiscoveredHost struct {
 	OSVersion                string   `json:"os_version"`
 	HostIPAddresses          []string `json:"ip_addresses"`
+	HostIPAddressesNetmasks  []string `json:"ip_addresses_netmasks"`
 	HostName                 string   `json:"hostname"`
 	CPUCount                 int      `json:"cpu_count"`
 	SocketCount              int      `json:"socket_count"`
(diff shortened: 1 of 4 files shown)


CDimonaco (approved):

EMaksy (commented):
Not a go expert but lgtm!

arbulu89 (commented):
It looks surprising to me that we are sending duplicated info now.
Did you consider updating the `ip_addresses` field itself, or sending only the masks in the new field?
I guess that this is because backward compatibility. Maybe we could handle that in the frontend somehow...

rtorrero:
> It looks surprising to me that we are sending duplicated info now. Did you consider updating the `ip_addresses` field itself, or sending only the masks in the new field? I guess that this is because backward compatibility. Maybe we could handle that in the frontend somehow...

Yes, I decided to go the safe way and split everything to avoid backwards compatibility issues. If we don't like this approach, we could include the netmask in the current `ip_addresses` discovery field and change the policy code to split in the host read model into `ip_addresses` and `ip_addresses_netmasks` there. The downside here is that we add a bit more complexity in the parsing. For me, a bit of duplication won't hurt too much, but I'm open to changing it :+1:

arbulu89:
Superseded by https://github.com/trento-project/agent/pull/346

PR #346, opened 2024-07-17
Discover host ip address netmasks

arbulu89:
# Description

Discover host netmasks and send them in the` netmasks` field. Each value corresponds the value in the same position for `ip_addresses`.

I'm aware that this "position related" usage is not optimal, and actually, in the 1st commit I implemented a more complex struct to send complete network interfaces. But this makes compatibility between old web/new agent, new web/old agent more difficult, so I sticked to simplest solution.

If for some reason, in the future we need more information about network interfaces, I finally think that we will need something more complex implementation that eventually will require some upgrade in api versioning or even major version of agent/web.

## How was this tested?

Partially tested...
We don't have any code testing host discoveries, and in fact, testing `network` related things is pretty  difficult.
Basically the serialization as json is the unique thing that it is tested.
In any case, the used functions are not changed, just the composed final code, so as everything was working now, i expect that it will continue working

### DIFF
--- a/test/fixtures/discovery/host/expected_published_host_discovery.json
+++ b/test/fixtures/discovery/host/expected_published_host_discovery.json
@@ -1,19 +1,49 @@
 {
-    "agent_id": "779cdd70-e9e2-58ca-b18a-bf3eb3f71244",
-    "discovery_type": "host_discovery",
-    "payload": {
-        "os_version": "15-SP2",
-        "ip_addresses": [
-            "10.1.1.4",
-            "10.1.1.5",
-            "10.1.1.6"
-        ],
-        "hostname": "thehostnamewherethediscoveryhappened",
-        "cpu_count": 2,
-        "socket_count": 1,
-        "total_memory_mb": 4096,
-        "agent_version": "trento-agent-version",
-        "installation_source": "Community",
-        "fully_qualified_domain_name": "com.example.trento.host"
-    }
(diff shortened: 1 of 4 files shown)


test/fixtures/discovery/host/expected_published_host_discovery.json
```diff
@@ -1,19 +1,24 @@
...
+  "discovery_type": "host_discovery",
+  "payload": {
+    "os_version": "15-SP2",
+    "ip_addresses": [
+      "10.1.1.4",
+      "10.1.1.5",
+      "10.1.1.6"
+    ],
+    "netmasks": [
+      24,
```
EMaksy, line 12: The only thing what concerns me is we must just think about using always the correct index in trento web :)

EMaksy (approved):
LGTM!

test/fixtures/discovery/host/expected_published_host_discovery.json
arbulu89: As the length of ip addresses and netmasks is always the same, you can use the standard `zip` and `unzip` functions to combine them pretty easily

EMaksy: Oh, today I learned about lodash [zipWith](https://lodash.com/docs/4.17.15#zipWith). Okay, this is cool!
Then never mind.

CDimonaco (approved):

PR #360, opened 2024-12-27
Context propagation in facts gathering flow

balanza:
# Description

Refactor all gatherers to accept a `Context` as the first argument.  These changes are preparatory and will enable parallel and concurrent gatherers to implement cancellation on failure (when running several gatherers at a time, we can kill all processes as soon as one reports failure).

So far, changes only affect interfaces and do not introduce behavior changes.

Plugins' interfaces are not changed as there is no seamless way to pass context to a plugin. This is inherited from the `go-plugin` architecture which is network-based.

## Changes
* the `FactGatherer.Gather` signature now requires a `Context` as the first argument (https://github.com/trento-project/agent/blob/34a793216907efc8330a0309f2d573db7e4c3a2a/internal/factsengine/gatherers/gatherer.go#L17)
* the `gather` command passes its context to the Gather function (https://github.com/trento-project/agent/blob/34a793216907efc8330a0309f2d573db7e4c3a2a/cmd/facts.go#L111)
* a cancellable group context is passed when receiving a gather request in the gathering flow (https://github.com/trento-project/agent/blob/34a793216907efc8330a0309f2d573db7e4c3a2a/internal/factsengine/gathering.go#L53)
* all gathers ignore the received context so far


cmd/facts.go
```diff
@@ -108,7 +108,7 @@ func gather(*cobra.Command, []string) {
 		},
 	}
 
-	value, err := g.Gather(factRequest)
+	value, err := g.Gather(cmd.Context(), factRequest)
```
arbulu89, line 111: Maybe I bit out of context, but we don't have any proper canceling on this command.
Could we maybe add something like this (copied from the `start` command):
```
signals := make(chan os.Signal, 1)
signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)

ctx, ctxCancel := context.WithCancel(context.Background())
// or
// ctx, ctxCancel := context.WithCancel(cmd.Context())

go func(ctx context.Context) {
	quit := <-signals
	log.Infof("Caught %s signal!", quit)

	log.Info("Stopping fact gathering...")
	ctxCancel()
}(ctx)

...

value, err := g.Gather(ctx, factRequest)
```

internal/factsengine/gatherers/_todo/crmmon_test.go
```diff
@@ -51,7 +52,7 @@ func (suite *CrmMonTestSuite) TestCrmMonGather() {
 		},
 	}
 
-	factResults, err := p.Gather(factRequests)
+	factResults, err := p.Gather(context.Background(), actRequests)
```
arbulu89, line 55: typo? should be `factRequests`.
It is not executed, but better to have it right hehe

internal/factsengine/gathering.go
```diff
@@ -24,7 +26,7 @@ func gatherFacts(
 	}
 	groupedFactsRequest := groupFactsRequestByGatherer(agentFacts)
 	factsCh := make(chan []entities.Fact, len(groupedFactsRequest.FactRequests))
-	g := new(errgroup.Group)
+	g, groupCtx := errgroup.WithContext(context.Background())
```
arbulu89, line 29: We should get this context still from above.
The context we use everywhere should be the one coming in the start of the agent (`a.Start(ctx)`) in the `start.go` file.
Otherwise, we don't use the same context.
I know that we need to pass this context in many many functions, but I think it is worthy. Just put `ctx context.Context` in a bunch of functions.
What do you think?

arbulu89 (changes requested):
Hey @balanza 
Looks mostly good.
Just two things. 
- I think we should use the context coming in `a.Start(ctx)` and pass it until reaching the gathering code
- We are passing this context because we are already using it in some gatherers. Could we use the passed context instead of creating a new one in the gatherer itself?
The gatherers that should be changed are: `ascsers_cluster`, `sapcontrol`, `systemd` and `systemd_v2` (just control+find hehe). I would keep the context usage in the `NewDefaultSystemDGathererV2` by now, as It would fail during initialization (we can pass it from above as well)

cmd/facts.go
arbulu89: I see this is coming in the next PR. We can discuss there.

balanza: Indeed. 

I purposely avoid introducing behavior changes in this PR, as it's already fat.

internal/factsengine/gathering.go
balanza: I'm not sure the context to use is the one relative to the `Start` command. This context is scoped on each gathering request (coming from rabbitmq).

As I intended it, we want all routines in the same request to terminate on failure. If we use the root context, we will have the whole agent process to fail.

arbulu89: > This context is scoped on each gathering request (coming from rabbitmq).

And it would continue this way, right? We would just be creating a new grouped context that can still be cancelled from above.

> If we use the root context, we will have the whole agent process to fail.
Is this true? 
I wouldn't expect the parent context being cancelled if some process in the group fails.
Maybe I'm wrong.

But what I would want is that if the parent context is cancelled (we stop trento agent), the gathering is cancelled and the execution stops, without leaving open traces or processes running.

If we create a new empty context, the stopping of the agent is not propagated to the gathering, and the gatherers that use contexts neither.

balanza: You're right, I didn't consider that scenario. I'll come up with a solution. Thanks!

balanza:
@arbulu89 some updates from the latest changes:
* now stopping the agent propagates the cancel signal to all the active gatherers (see https://github.com/trento-project/agent/pull/360/commits/3492b4d90fea66f47d27826130702c1ad3d9642f)
* same for the fact commands (see https://github.com/trento-project/agent/pull/360/commits/985b4ec605b4178b4e6fc70c447d7e08f93686f2)
* regardless of what I said in the comment before, one failing gatherer does not trigger cancellation on the others in its batch; this is because the actual logic is to collect both errors and results altogether.

internal/factsengine/policy.go
```diff
@@ -13,14 +14,22 @@ const (
 	FactsGatheringRequested = "Trento.Checks.V1.FactsGatheringRequested"
 )
 
-func (c *FactsEngine) handleEvent(_ string, request []byte) error {
+type eventHandler func(name string, event []byte) error
```
arbulu89, line 17: question: why this new `makeEventHandler`?
was not enough to change the signature of of the `Listen` function in the `adapter.go` file to use the context?
I don't know, I find it more confusing than before.
Now, if we have other functions listening, the will need to copy this, which makes things more opaque in my opinion

internal/factsengine/gathering.go
```diff
@@ -60,10 +72,12 @@ func gatherFacts(
 			}
 
 			return nil
+
 		})
 	}
 
-	if err := g.Wait(); err != nil {
+	err := g.Wait()
+	if err != nil {
```
arbulu89, line 80: nitpick: why do you remove the one liner? doesn't it look better?

internal/factsengine/gathering.go
```diff
@@ -48,7 +52,15 @@ func gatherFacts(
 		g.Go(func() error {
 			var gatheringError *entities.FactGatheringError
 
-			newFacts, err := gatherer.Gather(factsRequest)
+			newFacts, err := gatherer.Gather(ctx, factsRequest)
+
+			select {
```
arbulu89, line 57: question: here we are checking if the context is cancelled, right? At the end we are not waiting for it.
I think having this is equivalent:
```
newFacts, err := gatherer.Gather(groupCtx, factsRequest)
ctxErr := ctx.Err()
switch {
case ctxErr != nil:
    return ctxErr;
case err == nil:
    factsCh <- newFacts
...

// Or simply above the `switch`
if err := ctx.Err(); err != nil {
    return err
}
```
`ctx.Err` will return not nil if it is cancelled. Just to reuse the `switch`

arbulu89 (approved):
@balanza I wrote some "subjective" comments.
All good in any case.
Please don't forget using the `ctx` passed to the gatherers I mentioned in a previous comment before merging. Otherwise, all this is useless hehe

cmd/facts.go
```diff
@@ -108,7 +113,27 @@ func gather(*cobra.Command, []string) {
...
+		<-signals
+		log.Info("Caught signal!")
+		cancelled = true
+		cancel()
+
+	}()
+
+	value, err := g.Gather(ctx, factRequest)
+
+	if cancelled {
```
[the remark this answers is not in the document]
CDimonaco: Good catch we can remove the second cleanup plugins, the defer gets called in any case

internal/factsengine/gatherers/rcp_plugin.go
```diff
@@ -50,5 +52,17 @@ func (l *RPCPluginLoader) Load(pluginPath string) (FactGatherer, error) {
...
+	}
+
+	return p, nil
+}
+
+type PluggedGatherer struct {
+	plugin plugininterface.Gatherer
+}
+
+func (g *PluggedGatherer) Gather(_ context.Context, factsRequests []entities.FactRequest) ([]entities.Fact, error) {
```
CDimonaco, line 66: Nice 🤙🏻

internal/factsengine/gathering.go
CDimonaco: Considering the select is blocking this could be another valid approach

CDimonaco (approved):
LGTM

I left some comments, nitpicks

cmd/facts.go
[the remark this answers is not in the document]
arbulu89: Okey, as long as we don't call the cleanup twice I'm fine

internal/factsengine/gatherers/_todo/crmmon_test.go
```diff
@@ -90,7 +91,7 @@ func (suite *CrmMonTestSuite) TestCrmMonGatherCmdNotFound() {
 		},
 	}
 
-	_, err := p.Gather(factRequests)
+	_, err := p.Gather(context.Background(), actRequests)
```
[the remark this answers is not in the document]
balanza: fixed in https://github.com/trento-project/agent/pull/360/commits/1d7198c4eedddd271b6b626afd65b23a03f5a745

internal/factsengine/gathering.go
balanza: Unintentional, restored in https://github.com/trento-project/agent/pull/360/commits/c66e25b4bf909fb4aebf7830690e27b186833970

internal/factsengine/gathering.go
balanza: Done in https://github.com/trento-project/agent/pull/360/commits/ab5a951135e076bc44c808f9b1d907c0e2a7d420

internal/factsengine/policy.go
balanza: Because the listener in `FactsEngine.Listen` accepts a handling function with the signature `func(name string, event []byte) error`; and also, the context is coming from the scope not from the single event.

cmd/facts.go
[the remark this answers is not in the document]
balanza: Looking twice: the `defer` won't happen after a `log.Fatal` so the cleanup won't happen twice. Not my favourite code design, though.

internal/factsengine/policy.go
arbulu89: Yes, of course, but you can change the signatures they way you prefer. That's not an issue
```
Listen(ctx context.Context, queue, exchange, routingKey string, handle func(ctx context.Context, contentType string, message []byte) error) error
```

cmd/facts.go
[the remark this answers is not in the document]
arbulu89: mmm, i don't get it. You are doing this in the main thread, and you have a `return` afterwards. Where is the `log.Fatal` happening when you cancel?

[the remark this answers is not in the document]
balanza: I'm referring to the `cleanupAndFatal` function, I removed the other occurrence in https://github.com/trento-project/agent/pull/360/commits/055f17694b0ea4be8d0830642af22f52fe343efb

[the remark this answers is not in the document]
arbulu89: Yeah...
That was an ugly trick we implemented to always cleanup.
I'm not against changing it if we have a better alternative

internal/factsengine/policy.go
balanza: Idk, because then the `ctx` should be passed by `RabbitMQAdapter.Listen` which is not aware of it

arbulu89: I personally would pass it.
Up to you @balanza , I will respect the decision you take :slightly_smiling_face:

PR #362, opened 2025-01-08
Add tests on built binaries

balanza:
# Description

Introduce [BATS](https://github.com/bats-core/bats-core) to enable testing on built binaries, simulating the normal usage.

* add `bats` to the development tools
* add `*.bats` tests suites
* add `make test-build` command that run all the bats tests
* add `make build-plugin-examples` command that build the plugin examples into `./build/plutin-examples`
* add the commands to the CI

Furthermore, the build destination directory has been changed to `build/$(go mod GOARCH)`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ LDFLAGS = -X github.com/trento-project/agent/version.Version="$(VERSION)"
 LDFLAGS := $(LDFLAGS) -X github.com/trento-project/agent/version.InstallationSource="$(INSTALLATIONSOURCE)"
 ARCHS ?= amd64 arm64 ppc64le s390x
 DEBUG ?= 0
+BUILD_DIR ?= ./build
 
 ifeq ($(DEBUG), 0)
 	LDFLAGS += -s -w
@@ -18,7 +19,11 @@ default: clean mod-tidy fmt vet-check test build
 .PHONY: build
 build: agent
 agent:
-	$(GO_BUILD) -o trento-agent
+	$(GO_BUILD) -o $(BUILD_DIR)/trento-agent
+
+.PHONY: build-plugin-examples
+build-plugin-examples:
+	$(GO_BUILD) -o $(BUILD_DIR)/plugin_examples/dummy ./plugin_examples/dummy.go
 
 .PHONY: cross-compiled $(ARCHS)
 cross-compiled: $(ARCHS)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -4,6 +4,7 @@ LDFLAGS = -X github.com/trento-project/agent/version.Version="$(VERSION)"
 LDFLAGS := $(LDFLAGS) -X github.com/trento-project/agent/version.InstallationSource="$(INSTALLATIONSOURCE)"
 ARCHS ?= amd64 arm64 ppc64le s390x
 DEBUG ?= 0
+BUILD_DIR ?= ./build
```
arbulu89, line 7: This new default folder looks dangerous.
Many of our package generation pipelines (rpm creation spec files for example) depend on having the `trento-agent` binary in the root folder

arbulu89 (commented):
I like the idea!
Just be careful changing the default folder where the binaries are created.
Our packaging depends on having `trento-agent` binary in the root folder.
We would need to check and change all of this things.

Makefile
balanza: I'll try to fix everything to have all the artifacts under `./build`, it seems more tidy to me. If I cannot, I just revert to the current behavior

balanza: In the end, the release pipelines do not rely on the `make build` command, but use the `make cross-compile` instead, which creates an artifact for each supported architecture; The artifacts are placed into `build/<arch>` folders. 

Thanks to @CDimonaco's suggestion I opted for using the same convention for the development build and test.
So `make build` creates the artifact into the `build/${GOARCH}` folder, depending on the current system; so does `make build-plugin-examples` as well.

I think it's safe now.

arbulu89: @balanza The `rpm` creation does depend on the `make build` command and the output it generates: https://github.com/trento-project/agent/blob/main/packaging/suse/trento-agent.spec#L52

balanza: You are right, I forgot that part. I propose this to make it compatible: https://github.com/trento-project/agent/pull/362/commits/7e5320638f6568dde8f535a8447d5ef499685e0a

packaging/suse/trento-agent.spec
```diff
@@ -49,7 +49,7 @@ Trento agents are client-side processes responsible for the automatic discovery
 %define shortname agent
 
 %build
-VERSION=%{version} INSTALLATIONSOURCE=Suse make build
+VERSION=%{version} INSTALLATIONSOURCE=Suse make build BUILD_OUTPUT="./trento-agent"
```
arbulu89, line 52: nitpick: I would put `BUILD_OPTION` after `INSTALLATION_SOURCE` and before `make build`, so we have `envs command` structure

arbulu89 (approved):
Brilliant!

PR #371, opened 2025-01-13
Add healthcheck to prevent flacky tests

balanza:
When running `make test` in the CI, the rabbitmq service may eventually not be ready to accept connections. 
Adding a healthcheck command before running the actual test script.


balanza:
It seems it's not really working 😅

CDimonaco (approved):
image

PR #379, opened 2025-01-22
Implement context cancellation for command based gatherers

balanza:
# Description

Implement context cancellation for all gatherers that execute commands on the host machine:
* ascsers
* cibadmin
* corosync
* dispwork 
* mountinfo 
* packageversion 
* saphostctrl 
* sbddump 
* sysctl
* verifypassword

### DIFF
--- a/internal/factsengine/gatherers/ascsers_cluster.go
+++ b/internal/factsengine/gatherers/ascsers_cluster.go
@@ -75,13 +75,11 @@ func (g *AscsErsClusterGatherer) SetCache(cache *factscache.FactsCache) {
 }
 
 func (g *AscsErsClusterGatherer) Gather(
-	_ context.Context,
+	ctx context.Context,
 	factsRequests []entities.FactRequest,
 ) ([]entities.Fact, error) {
 	log.Infof("Starting %s facts gathering process", AscsErsClusterGathererName)
 	var cibdata cib.Root
-
-	ctx := context.Background()
 
 	content, err := factscache.GetOrUpdate(
 		g.cache,
--- a/internal/factsengine/gatherers/ascsers_cluster.go
+++ b/internal/factsengine/gatherers/ascsers_cluster.go
@@ -86,7 +86,7 @@ func (g *AscsErsClusterGatherer) Gather(
 	content, err := factscache.GetOrUpdate(
 		g.cache,
 		CibAdminGathererCache,
-		memoizeCibAdmin,
+		makeMemoizeCibAdmin(ctx),
 		g.executor,
 	)
 
--- a/internal/factsengine/gatherers/ascsers_cluster_test.go
+++ b/internal/factsengine/gatherers/ascsers_cluster_test.go
@@ -293,7 +293,7 @@ func (suite *AscsErsClusterTestSuite) TestAscsErskGathererContextCancelled() {
 	cancel()
 
 	suite.mockExecutor.
-		On("ExecContext", mock.Anything, "cibadmin", "--query", "--local").
+		On("ExecContext", ctx, "cibadmin", "--query", "--local").
 		Return(nil, ctx.Err())
 
 	c := gatherers.NewAscsErsClusterGatherer(suite.mockExecutor, suite.webService, nil)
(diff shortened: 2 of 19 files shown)


internal/factsengine/gatherers/ascsers_cluster_test.go
```diff
@@ -286,3 +287,26 @@ func (suite *AscsErsClusterTestSuite) TestAscsErsClusterGather() {
 	}
 	suite.ElementsMatch(expectedEntries, entries)
 }
+
+func (suite *AscsErsClusterTestSuite) TestAscsErskGathererContextCancelled() {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	suite.mockExecutor.
+		On("ExecContext", mock.Anything, "cibadmin", "--query", "--local").
```
arbulu89, line 296: Did you try putting `ctx` instead of `mock.Anything`?
Maybe it validates that the function is called with the expected argument.
I don't know if it works though.

arbulu89 (commented):
Hey @balanza 
Good job. Just some minor changes to do.

Besides that:
A question about the new test cases.
Why do you mock the execution here? you could test a real cancellation as well, not using a mock. At the end, it is cancelled and it should return the same error.
(for example, this has hidden the fact that you didn't use the `ctx` coming from `Gather` in the `ascsers_cluster.go`

If you think that is already well tested and you prefer to mock is OK.

nitpick:I think we could improve a bit them changing `suite.Error(err)` to see that this actually has a cancellation error rather than any possible error.

internal/factsengine/gatherers/ascsers_cluster.go
```diff
@@ -86,7 +86,7 @@ func (g *AscsErsClusterGatherer) Gather(
 	content, err := factscache.GetOrUpdate(
 		g.cache,
 		CibAdminGathererCache,
-		memoizeCibAdmin,
+		makeMemoizeCibAdmin(ctx),
```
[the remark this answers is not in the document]
balanza: Definitely yes! Good catch, thanks. Done it in https://github.com/trento-project/agent/pull/379/commits/bb7d346f976c4c152f8008f6d74952f3c373afc7

It actually exposed the weakness in the proposed tests: it does not test that you're passing the correct context to the executor (let's discuss it below)

internal/factsengine/gatherers/ascsers_cluster_test.go
balanza: It works and it's actually brilliant! I refactored all tests like that in https://github.com/trento-project/agent/pull/379/commits/596152435a42cc5abe3aaf86f435937ca1e89013

It would have spotted the bug in https://github.com/trento-project/agent/pull/379#discussion_r1926499834
You made my day

arbulu89 (dismissed):
Great! :green_heart: 
You can merge once the `Ersk` typo is fixed!

balanza:
> A question about the new test cases.
Why do you mock the execution here? you could test a real cancellation as well, not using a mock. At the end, it is cancelled and it should return the same error.

You're right. In the end, command cancellation (and cleanup) is covered by tests on the command executor itself.

I refactor all tests in https://github.com/trento-project/agent/pull/379/commits/b064d323f9d672b12a09452156047bfee4f328bf and they work smoothly using the real executor. _The less we mock the better_, here it comes with a big BUT:

tests will succeed even if `ExecCommand` isn't hit. For example, imagine the case in which the `Gather` implementation has an early return on `ctx.Err()`. 
By doing this, we are _theoretically_ missing the scenario in which the command has been started and then cancelled.

Anyway, it's nitpick and it's not our case so far. I like what we have now, I just wanted to highlight a possible drawback. 

If you are fine with that, we can merge.

arbulu89 (approved):
@balanza That's ok. 
I don't see why we should be checking `ctx.Err` before doing anything. It looks like something we shouldn't do.
At the end, the context we pass is to be used by system calls (commands, api calls, etc), so we can cancel them (even what we return is not that critical).

I think we cover all of our real cases

PD: if we have some early `ctx.Err` usage in the code, we should aim to a different test format as well

PR #383, opened 2025-01-24
Add context handling to coosyncconf gather

balanza:
Following https://github.com/trento-project/agent/pull/376

In addition, two helper functions are added to deal with channel and file reads

### DIFF
--- a/internal/factsengine/gatherers/corosyncconf.go
+++ b/internal/factsengine/gatherers/corosyncconf.go
@@ -1,15 +1,14 @@
 package gatherers
 
 import (
-	"bufio"
 	"context"
 	"fmt"
-	"os"
 	"regexp"
 
-	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
 	"github.com/trento-project/agent/pkg/factsengine/entities"
+	"github.com/trento-project/agent/pkg/utils"
 )
 
 const (
@@ -51,13 +50,13 @@ func NewCorosyncConfGatherer(configFile string) *CorosyncConfGatherer {
 }
 
 func (s *CorosyncConfGatherer) Gather(
-	_ context.Context,
+	ctx context.Context,
 	factsRequests []entities.FactRequest,
 ) ([]entities.Fact, error) {
 	facts := []entities.Fact{}
 	log.Infof("Starting corosync.conf file facts gathering process")
 
-	corosyncConfile, err := readCorosyncConfFileByLines(s.configFile)
+	corosyncConfile, err := utils.ScanFileContext(ctx, afero.NewOsFs(), s.configFile)
 	if err != nil {
 		return nil, CorosyncConfFileError.Wrap(err.Error())
 	}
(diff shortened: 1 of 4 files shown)


internal/factsengine/gatherers/corosyncconf.go
```diff
@@ -83,26 +82,10 @@ func (s *CorosyncConfGatherer) Gather(
...
-
-	defer corosyncConfFile.Close()
-
-	fileScanner := bufio.NewScanner(corosyncConfFile)
-	fileScanner.Split(bufio.ScanLines)
-	var fileLines []string
-
-	for fileScanner.Scan() {
-		fileLines = append(fileLines, fileScanner.Text())
+	if ctx.Err() != nil {
```
arbulu89, line 85: Do we need to check this?
I mean, `utils.ScanFileContext`is already returning an error in this case.
I wouldn't worry to check this thinking that between the scan file and this small piece of code we could have the cancellation.
If we are going to do so in all the gatherers (I think it is an overkill) I think we should do it in the `gathering.go` code.
There we already have a `switch` after running the `Gather`, so we could check there.
The purpose of thie context is to not block rather than notifying an error

internal/factsengine/gatherers/corosyncconf_test.go
```diff
@@ -212,7 +212,7 @@ func (suite *CorosyncConfTestSuite) TestCorosyncConfFileNotExists() {
 	factsGathered, err := c.Gather(context.Background(), factsRequest)
 
 	expectedError := &entities.FactGatheringError{
-		Message: "error reading corosync.conf file: could not open corosync.conf file: " +
+		Message: "error reading corosync.conf file: could not open not_found: " +
```
arbulu89, line 215: nitpick: Just with this output, it looks like we are not piping properly the error messages

pkg/utils/channels.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "could not open "+filePath)
+	}
+
+	defer file.Close()
+
+	errChan := make(chan error, 1)
+	contentChan := make(chan []string)
+
+	go func() {
```
arbulu89, line 41: question: Why do we need a thread here?
Having the `for` and `select` in the main thread is not enough?
At the end, you are staring the thread and waiting right away blocking the function.

arbulu89 (commented):
Hey @balanza 
I will play the devils advocate.
Which are the chances and consequences in time to get a cancellation request during this files reading? (considering the file might be around 50 lines).
Even if you catch that exact same point, how long this would block the cancellation?
Let's not forget that the propagation of the context is mostly to avoid a blocked state during on this, more than error reporting (if we are going to cancel, we don't care that much about the error).

In the commands and other operations had more sense, as we can have long going command executions, but reading files, which most of the cases are small.

I have the feeling that this is as overcomplicated solution for a problem we don't have.
What do you think?

arbulu89:
> @arbulu89 I have the same feeling. I made the PR also to raise any concerns about whether it's needed or not.
> 
> I think we can get rid of the code and keep the test, to ensure the interfaces are confirmed. What do you think?

I'm fine with that.
If we reuse a lot this file scanning, I'm not against having a utils function (even with the context cancelling), but I think we could have it without new threads or channels, just returning from the function when the `ctx.Done()` is done with the error.

PR #390, opened 2025-02-05
Handle context cancellation in systemd gatherer 

balanza:
Following https://github.com/trento-project/agent/pull/376


internal/factsengine/gatherers/systemd.go
```diff
@@ -63,43 +63,58 @@ func NewSystemDGatherer(conn DbusConnector, initialized bool) *SystemDGatherer {
...
-		if len(factReq.Argument) == 0 {
-			continue
-		}
-		services = append(services, completeServiceName(factReq.Argument))
-	}
+	results := make(chan []entities.Fact)
+	errors := make(chan error)
 
-	ctx := context.Background()
+	go func() {
```
arbulu89, line 76: question: why do we need a thread?
I would suppose that the `ListUnitsByNamesContext` function returns immediately when the context is cancelled. If that's the case we can simply return the error it returns.
We basically want to return when the context is cancelled, we don't really care even which error we return, we are simply cancelling the whole process.

At omst you can check if the context is done after running the `ListUnitsByNamesContext` if you want to have a different error

arbulu89 (commented):
Hey @balanza 
I dropped a question.
I think simply replacing the incoming `ctx` by the one we create in the gatherer would do what we want

internal/factsengine/gatherers/systemd.go
balanza: It's a bit overkill, indeed. The refactoring _per se_ has two benefits:
1. don't wait for `ListUnitsByNamesContext` to finish before closing the gathering process
2. have an explicit function response

If you think both are of not much value, we can rollback

arbulu89: > don't wait for ListUnitsByNamesContext to finish before closing the gathering process

I think we would wait the same either way. I expect `ListUnitsByNamesContext` returning right away when the context we pass is cancelled, and we return from that function if this happens. If `ListUnitsByNamesContext` doesn't work that way it is a bit of a bummer, as we don't even need to pass the context to it (it is not doing anything).
In fact, we can enter in a race condition and have a different result depending on the `ctx.Done` of the `ListUnitsByNamesContext` or the one you are using in the `select` reacts first.

balanza: Makes sense

arbulu89 (approved):
Good!
Some nitpick comments

internal/factsengine/gatherers/systemd_test.go
```diff
@@ -178,3 +179,56 @@ func (suite *SystemDTestSuite) TestSystemDGatherError() {
...
+
+	factsRequest := []entities.FactRequest{
+		{
+			Name:     "no_argument_fact",
+			Gatherer: "systemd",
+			CheckID:  "check1",
+		}}
+
+	go func() {
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
balanza: Done in https://github.com/trento-project/agent/pull/390/commits/fa0b89ee7958d117d55a12c64acb8049786662c7

internal/factsengine/gatherers/systemd_test.go
```diff
@@ -178,3 +179,56 @@ func (suite *SystemDTestSuite) TestSystemDGatherError() {
...
 		"error getting unit states: error listing")
 }
+
+func (suite *SystemDTestSuite) TestSystemDContextCancelled() {
+	mockConnector := mocks.NewDbusConnector(suite.T())
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
+	mockConnector.On("ListUnitsByNamesContext", mock.Anything, []string{}).Return(
```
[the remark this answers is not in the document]
balanza: For the first test is fine to use the real connector, I did it in https://github.com/trento-project/agent/pull/390/commits/8cb175d53114e375d0865924020c1ca719512dc1

The second test I keep using the mock so I can simulate a cancellation occurring during the command execution

PR #391, opened 2025-02-05
Update agent to golang 1.23

balanza:
This PR updates the agent code/building infrastructure to be built with golang 1.23.

Version 1.23 is available in our build system at https://build.opensuse.org/package/show/devel:languages:go/go1.23

Some adjustments were needed in the code, please follow the comments below for the discussion.

### DIFF
--- a/cmd/facts.go
+++ b/cmd/facts.go
@@ -190,6 +190,6 @@ func list(*cobra.Command, []string) {
 	log.Printf("Available gatherers:")
 
 	for _, g := range gatherers {
-		log.Printf(g)
+		log.Printf("%s", g)
 	}
 }
--- a/cmd/facts.go
+++ b/cmd/facts.go
@@ -190,6 +190,6 @@ func list(*cobra.Command, []string) {
 	log.Printf("Available gatherers:")
 
 	for _, g := range gatherers {
-		log.Printf("%s", g)
+		log.Print(g)
 	}
 }
--- a/internal/core/hosts/discovered_host.go
+++ b/internal/core/hosts/discovered_host.go
@@ -7,7 +7,7 @@ type DiscoveredHost struct {
 	HostName                 string            `json:"hostname"`
 	CPUCount                 int               `json:"cpu_count"`
 	SocketCount              int               `json:"socket_count"`
-	TotalMemoryMB            int               `json:"total_memory_mb"`
+	TotalMemoryMB            uint64            `json:"total_memory_mb"`
 	AgentVersion             string            `json:"agent_version"`
 	InstallationSource       string            `json:"installation_source"`
 	FullyQualifiedDomainName *string           `json:"fully_qualified_domain_name,omitempty"`
--- a/internal/discovery/host.go
+++ b/internal/discovery/host.go
@@ -170,12 +170,12 @@ func getOSVersion() string {
 	return infoStat.PlatformVersion
 }
 
-func getTotalMemoryMB() int {
+func getTotalMemoryMB() uint64 {
 	v, err := mem.VirtualMemory()
 	if err != nil {
 		log.Errorf("Error while getting memory info: %s", err)
 	}
-	return int(v.Total) / 1024 / 1024
+	return v.Total / 1024 / 1024
 }
 
 func getLogicalCPUs() int {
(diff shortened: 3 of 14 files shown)


internal/core/hosts/discovered_host.go
```diff
@@ -7,7 +7,7 @@ type DiscoveredHost struct {
 	HostName                 string            `json:"hostname"`
 	CPUCount                 int               `json:"cpu_count"`
 	SocketCount              int               `json:"socket_count"`
-	TotalMemoryMB            int               `json:"total_memory_mb"`
+	TotalMemoryMB            uint64            `json:"total_memory_mb"`
```
balanza, line 10: This change is needed to comply with the new signature of the function `getTotalMemoryMB` (see comment below).

The change is made according to the following rationales:
1. We only support 64-bit architectures, thus `int` is always the same as `int64`
2. `int`, `int64` and `uint64` have the same JSON representation 
3. `uint64` has the same precision loss as `int` when marshalling to JSON, as  their value cannot exceed 2⁵³−1

internal/discovery/host.go
```diff
@@ -170,12 +170,12 @@ func getOSVersion() string {
 	return infoStat.PlatformVersion
 }
 
-func getTotalMemoryMB() int {
+func getTotalMemoryMB() uint64 {
```
balanza, line 173: This change is needed to fix the lint error
```
G115: integer overflow conversion uint64 -> int64
```

The change is considered safe because:
1. We only support 64-bit architectures
2. There is no negative memory byte count

Alternatively, to leave the signature untouched we would have to prove that the runtime value is inside the `int` range for the running architecture; is not trivial to decide what to do if it does not (panic? return default?)

internal/factsengine/gatherers/fstab.go
```diff
@@ -60,7 +62,13 @@ func (f *FstabGatherer) Gather(ctx context.Context, factsRequests []entities.Fac
...
 	entries := []FstabEntry{}
 
-	for _, m := range mounts {
+	for i, m := range mounts {
+		if m.PassNo < 0 {
+			return nil, FstabFileDecodingError.Wrap("invalid check order for mount" + strconv.Itoa(i))
+		}
+		if m.Freq < 0 || m.Freq > math.MaxUint8 {
+			return nil, FstabFileDecodingError.Wrap("invalid backup frequency for mount" + strconv.Itoa(i))
+		}
```
balanza, line 71: The change is needed to fix the lint error 
```
G115: integer overflow conversion int -> uint8
G115: integer overflow conversion int -> uint64
```

The _guards_ on the runtime value are needed to prove to the linter that both values are in the expected range.

internal/factsengine/gatherers/groups.go
```diff
@@ -108,6 +108,9 @@ func parseGroupsFile(fileContent io.Reader) ([]GroupsEntry, error) {
 		if err != nil {
 			return nil, fmt.Errorf("could not convert group id %s to integer", values[2])
 		}
+		if groupID < 0 {
+			return nil, fmt.Errorf("group id %d is less than 0", groupID)
+		}
```
balanza, line 113: The change is needed to fix the lint error 
```
G115: integer overflow conversion int -> uint64
```

nelsonkopliku:
hey @balanza besides being available on opensuse repos, is go 1.23 available also on official sles (and sles4sap) repos? Cause when building the final artifact we're not gonna be using opensuse repos.

cmd/facts.go
```diff
@@ -190,6 +190,6 @@ func list(*cobra.Command, []string) {
 	log.Printf("Available gatherers:")
 
 	for _, g := range gatherers {
-		log.Printf(g)
+		log.Printf("%s", g)
```
arbulu89, line 193: You could change to `log.Print` as well

arbulu89 (approved):
Good!
I have checked and go1.23 is available in our repositories from SLES15SP3 and onward.
We don't really care if it is available in openSUSE if it is not internally (generally, the should match, but it is not always the case).

Remember to do the same upgrade in the contracts repository: https://github.com/trento-project/contracts

nelsonkopliku (approved):

PR #395, opened 2025-02-20
Add gatherer to retrieve facts from `global.ini` configuration

balanza:
# Description

Introduce the `ini_files` gatherer that retrieves information as facts from a white-listed set of well-known files.
The first implementation only supports `global.ini` files

## Key features
* it assumes `/usr/sap` as base installation path and and `/usr/sap/$SID/SYS/global/hdb/custom/config` as the configuration directory for every installed SAP system on the host;
* it retrieves the content of the `global.ini` file for every installed SAP system on the host 
* it expects every SAP system to have a valid `global.ini` file, otherwise an error fact is retrieved

## Example
Supposing that there is one SAP system on the host with a global configuration like
```ini
[communication]
internal_network = 10.23.1.128/26
listeninterface = .internal
[internal_hostname_resolution]
10.23.1.132 = hana-s1-db1
10.23.1.133 = hana-s1-db2
10.23.1.134 = hana-s1-db3
```

```sh
$ trento-agent facts gather --gatherer ini_files --argument global.ini
INFO[2025-02-20 23:15:10] loading plugins                              
INFO[2025-02-20 23:15:10] Starting ini_files facts gathering process   
INFO[2025-02-20 23:15:10] New SAP system installation found: S01       
INFO[2025-02-20 23:15:10] Gathered fact for "ini_files" with argument "global.ini": 
INFO[2025-02-20 23:15:10] Name: global.ini
Check ID: 

Value:

[
  #{
    "sid": "S01",
    "value": #{
      "communication": #{
        "internal_network": "10.23.1.128/26",
        "listeninterface": ".internal"
      },
      "internal_hostname_resolution": #{
        "10.23.1.132": "hana-s1-db1",
        "10.23.1.133": "hana-s1-db2",
        "10.23.1.134": "hana-s1-db3"
      }
    }
  }
]
```


arbulu89 (commented):
Hey @balanza 
Code aside, we are not returning to which SID belongs each ini file content, and I think it might be relevant. Maybe it might not be used in the checks, but if you have a list of SAP systems and getting a list of ini files without knowing to which system they belong looks "incomplete".
I think it is a tiny change that will come handy.
In the checks, if they don't need the SID, they can simply loop using the values of the map

balanza:
> Hey @balanza Code aside, we are not returning to which SID belongs each ini file content, and I think it might be relevant. Maybe it might not be used in the checks, but if you have a list of SAP systems and getting a list of ini files without knowing to which system they belong looks "incomplete". I think it is a tiny change that will come handy. In the checks, if they don't need the SID, they can simply loop using the values of the map

Done in https://github.com/trento-project/agent/pull/395/commits/73a71522eb36e9cb3a9ec879f71609281aa1926a

internal/factsengine/gatherers/ini_files.go
```diff
@@ -0,0 +1,164 @@
...
+		return entities.Fact{}, err
+	}
+	if len(sids) == 0 {
+		return entities.Fact{}, errors.New("no SAP system found")
+	}
+
+	values := entities.FactValueList{}
+
+	for _, sid := range sids {
+		path := globalIniPath(sid)
```
arbulu89, line 92: nitpick for the future: No need to refactor now, but I guess we will need to somehow duplicate this if we introduce new files to parse.
Maybe passing the function to get the path as param could work.
Just to not duplicate the whole function

internal/factsengine/gatherers/ini_files.go
```diff
@@ -0,0 +1,164 @@
...
+
+	cfg, err := ini.Load(content)
+	if err != nil {
+		return nil, errors.Wrap(err, "error loading ini file")
+	}
+
+	result := make(map[string]interface{})
+
+	for _, section := range cfg.Sections() {
+		sectionMap := make(map[string]interface{})
```
arbulu89, line 135: refactor: maybe you can simplify this as:
```
result := make(map[string]interface{})
for _, section := range cfg.Sections() {
    if section.Name() == ini.DefaultSection {
        for _, key := range section.Keys() {
            result[key.Name()] = key.String()
        }
        continue
    }

    sectionMap := make(map[string]interface{})
    for _, key := range section.Keys() {
        sectionMap[key.Name()] = key.String()
    }
    result[section.Name()] = sectionMap
}
```
We would remove one loop execution

test/cli/gatherers.bats
```diff
@@ -40,6 +40,55 @@ function teardown() {
...
+        -v "$BUILD_DIR/trento-agent":/usr/bin/trento-agent \
+        -v "$sandbox/$config_dir":"$config_dir" \
+        opensuse/leap:15.6 \
+        trento-agent facts gather --gatherer ini_files --argument global.ini
+
+    expected=$(cat <<EOF
+[
+  #{
+    "sid": "S01",
+    "value": #{
```
arbulu89, line 67: In the code and code tests you use `content` and `value` here.
How can this work? Am i missing something?

internal/factsengine/gatherers/ini_files_test.go
```diff
@@ -0,0 +1,287 @@
...
+	factRequests := []entities.FactRequest{
+		{
+			Name:     "any",
+			Gatherer: "ini_files",
+		},
+	}
+
+	_, err := c.Gather(context.Background(), factRequests)
+
+	suite.Error(err)
```
arbulu89, line 34: nitpick: I think we can improve the error checking with `suite.EqualError`. This is for all the tests returning errors

internal/factsengine/gatherers/ini_files_test.go
```diff
@@ -0,0 +1,287 @@
+package gatherers_test
+
+import (
+	"context"
+	"testing"
+
+	st "github.com/balanza/supertouch"
```
arbulu89, line 7: question: why did you choose to create your own project for this?
afero was not good enough?
I think we could have a tiny presentation about the  package to the rest of the team to avoid isolated knowledge

arbulu89 (approved):
Looks good!
I just wonder how the bats test is passing using `value` instead of `content`.
Now, the final step would be to document the gatherer in https://github.com/trento-project/wanda/blob/main/guides/gatherers.md

internal/factsengine/gatherers/ini_files_test.go
balanza: I wished `afero` could do it, but it does not afaik.

That's a utility I made for another project, I kinda missed it here. We may as well copy the files, it's not  rocket science

arbulu89: We already use `afero` to create on memory file systems. For example: https://github.com/trento-project/agent/blob/main/internal/factsengine/gatherers/sapservices_test.go#L39
Actually, we chose to use `afero` to make testing possible

internal/factsengine/gatherers/ini_files_test.go
```diff
@@ -0,0 +1,256 @@
...
+					Value: map[string]entities.FactValue{
+						"key2": &entities.FactValueString{Value: "value2"},
+					},
+				},
+			},
+		})
+
+}
+
+func (suite *IniFilesTestSuite) TestIniFilesGathererGlobalIniPartial() {
```
arbulu89, line 209: question: just curiosity. Why is this test sending a fact with an error?

internal/factsengine/gatherers/ini_files_test.go
balanza: Fair enough, I removed it in https://github.com/trento-project/agent/pull/395/commits/5a1d8cefbd7570b23e7e2da52a3193138081efd2

internal/factsengine/gatherers/ini_files_test.go
arbulu89: In fact, the error is `*entities.FactGatheringError)(nil)`. And it has a correct value.
I don't know if this test is really useful

balanza: I changed the test because what I wanted to prove is how code handles the case in which one of the parsing fails

test/cli/gatherers.bats
balanza: I removed the test because I didn't find a way to test the output in bash

arbulu89 (approved):
Perfect!

PR #404, opened 2025-03-04
Add `sudoers` gatherer

balanza:
# Description

Add a `sudoers` gatherer to retrieve information about the permission of users, in order to verify the compliance of the host configuration. The gatherer returns a parsed json with the output of the command
```sh
sudo -l -U <user>
```

The gatherer operates in two modes:
* when a user is specified, it fetches the information about the single user
* otherwise, it discovers the SAP users configured in the host and returns data for each user; SAP users are defined by convention using the lower-cased system id and the suffix `adm`

In both cases, the result is a list.

### Example

```sh
$ sudo -l -U prdadm

User prdadm may run the following commands on usr-vmhana01:
    (ALL) ALL
    (ALL) NOPASSWD: /usr/sbin/crm_attribute -n hana_prd_site_srHook_Site1 -v SOK -t crm_config -s SAPHanaSR, /usr/sbin/crm_attribute -n hana_prd_site_srHook_Site1 -v SFAIL -t crm_config -s
        SAPHanaSR, /usr/sbin/crm_attribute -n hana_prd_site_srHook_Site2 -v SOK -t crm_config -s SAPHanaSR, /usr/sbin/crm_attribute -n hana_prd_site_srHook_Site2 -v SFAIL -t crm_config -s
        SAPHanaSR
    (ALL) NOPASSWD: /usr/sbin/SAPHanaSR-hookHelper --case checkTakeover --sid\=prd

$ trento-agent facts gather --gatherer sudoers --argument prdadm

Value:

[
  #{
    "command": "ALL",
    "no_password": false,
    "run_as_group": "",
    "run_as_user": "ALL",
    "user": "prdadm"
  },
  #{
    "command": "/usr/sbin/crm_attribute -n hana_prd_site_srHook_Site1 -v SOK -t crm_config -s SAPHanaSR",
    "no_password": true,
    "run_as_group": "",
    "run_as_user": "ALL",
    "user": "prdadm"
  },
  #{
    "command": "/usr/sbin/crm_attribute -n hana_prd_site_srHook_Site1 -v SFAIL -t crm_config -s SAPHanaSR",
    "no_password": true,
    "run_as_group": "",
    "run_as_user": "ALL",
    "user": "prdadm"
  },
  #{
    "command": "/usr/sbin/crm_attribute -n hana_prd_site_srHook_Site2 -v SOK -t crm_config -s SAPHanaSR",
    "no_password": true,
    "run_as_group": "",
    "run_as_user": "ALL",
    "user": "prdadm"
  },
  #{
    "command": "/usr/sbin/crm_attribute -n hana_prd_site_srHook_Site2 -v SFAIL -t crm_config -s SAPHanaSR",
    "no_password": true,
    "run_as_group": "",
    "run_as_user": "ALL",
    "user": "prdadm"
  },
  #{
    "command": "/usr/sbin/SAPHanaSR-hookHelper --case checkTakeover --sid\\=prd",
    "no_password": true,
    "run_as_group": "",
    "run_as_user": "ALL",
    "user": "prdadm"
  }
]
```

### DIFF
--- a/internal/factsengine/gatherers/sudoers.go
+++ b/internal/factsengine/gatherers/sudoers.go
@@ -182,58 +182,61 @@ func (g *SudoersGatherer) parseUserPrivileges(output string) ([]privilegeEntry, 
 			continue // Skip the header line itself.
 		}
 
-		if inPrivilegesSection {
-			trimmed := strings.TrimSpace(line)
-			if trimmed == "" {
-				continue
-			}
-
-			// Attempt to extract all expected parts from the line.
-			matches := entryRegex.FindStringSubmatch(line)
-			if len(matches) == 0 {
-				// If the line does not match, it might be a continuation of the previous entry.
-				// Advanced parsing would be required to handle multi-line entries.
-				continue
-			}
-
-			// Matches:
-			// matches[1]: run-as specifier (e.g. "ALL : ALL" or "root")
-			// matches[2]: optional password flag (e.g. "NOPASSWD:")
-			// matches[3]: the commands string
-			runas := matches[1]
-			flag := matches[2]
-			commandsStr := matches[3]
-
-			var runAsUser, runAsGroup string
-			parts := strings.Split(runas, ":")
-			runAsUser = strings.TrimSpace(parts[0])
-			if len(parts) > 1 {
-				runAsGroup = strings.TrimSpace(parts[1])
-			}
-
-			noPassword := false
-			if flag == "NOPASSWD:" {
-				noPassword = true
-			}
-
-			// Commands might be a comma-separated list. Split and trim each command.
-			rawCommands := strings.Split(commandsStr, ",")
-			var commands []string
-			for _, cmd := range rawCommands {
-				trimmedCmd := strings.TrimSpace(cmd)
-				if trimmedCmd != "" {
-					commands = append(commands, trimmedCmd)
-				}
-			}
-
-			entry := privilegeEntry{
-				runAsUser:  runAsUser,
-				runAsGroup: runAsGroup,
-				noPassword: noPassword,
-				commands:   commands,
-			}
-			privileges = append(privileges, entry)
+		if !inPrivilegesSection {
+			continue // Skip lines before the privileges section.
 		}
+
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" {
+			continue
+		}
+
+		// Attempt to extract all expected parts from the line.
+		matches := entryRegex.FindStringSubmatch(line)
+		if len(matches) == 0 {
+			// If the line does not match, it might be a continuation of the previous entry.
+			// Advanced parsing would be required to hayndle multi-line entries.
+			continue
+		}
+
+		// Matches:
+		// matches[1]: run-as specifier (e.g. "ALL : ALL" or "root")
+		// matches[2]: optional password flag (e.g. "NOPASSWD:")
+		// matches[3]: the commands string
+		runas := matches[1]
+		flag := matches[2]
+		commandsStr := matches[3]
+
+		var runAsUser, runAsGroup string
+		parts := strings.Split(runas, ":")
+		runAsUser = strings.TrimSpace(parts[0])
+		if len(parts) > 1 {
+			runAsGroup = strings.TrimSpace(parts[1])
+		}
+
+		noPassword := false
+		if flag == "NOPASSWD:" {
+			noPassword = true
+		}
+
+		// Commands might be a comma-separated list. Split and trim each command.
+		rawCommands := strings.Split(commandsStr, ",")
+		var commands []string
+		for _, cmd := range rawCommands {
+			trimmedCmd := strings.TrimSpace(cmd)
+			if trimmedCmd != "" {
+				commands = append(commands, trimmedCmd)
+			}
+		}
+
+		entry := privilegeEntry{
+			runAsUser:  runAsUser,
+			runAsGroup: runAsGroup,
+			noPassword: noPassword,
+			commands:   commands,
+		}
+		privileges = append(privileges, entry)
+
 	}
 
 	if !inPrivilegesSection {
--- a/internal/factsengine/gatherers/sudoers.go
+++ b/internal/factsengine/gatherers/sudoers.go
@@ -175,7 +175,7 @@ func (g *SudoersGatherer) parseUserPrivileges(output string) ([]privilegeEntry, 
 	// Group 1 captures the run-as specifier inside parentheses.
 	// Group 2 optionally captures "NOPASSWD:" or "PASSWD:".
 	// Group 3 captures the commands and any arguments.
-	entryRegex := regexp.MustCompile(`^\s*\(([^)]+)\)\s*(?:(NOPASSWD:|PASSWD:)\s*)?(.*)$`)
+	entryRegex := regexp.MustCompile(`^\s*\(([^)]+)\)\s*(?:(([A-Z]+:)+)\s*)?(.*)$`)
 
 	for scanner.Scan() {
 		line := scanner.Text()
@@ -208,8 +208,8 @@ func (g *SudoersGatherer) parseUserPrivileges(output string) ([]privilegeEntry, 
 		// matches[2]: optional password flag (e.g. "NOPASSWD:")
 		// matches[3]: the commands string
 		runas := matches[1]
-		flag := matches[2]
-		commandsStr := matches[3]
+		flags := matches[2]
+		commandsStr := matches[4]
 
 		var runAsUser, runAsGroup string
 		parts := strings.Split(runas, ":")
@@ -218,7 +218,10 @@ func (g *SudoersGatherer) parseUserPrivileges(output string) ([]privilegeEntry, 
 			runAsGroup = strings.TrimSpace(parts[1])
 		}
 
-		noPassword := flag == "NOPASSWD:"
+		noPassword := false
+		for _, flag := range strings.Split(flags, ":") {
+			noPassword = noPassword || strings.TrimSpace(flag) == "NOPASSWD"
+		}
 
 		// Commands might be a comma-separated list. Split and trim each command.
 		rawCommands := strings.Split(commandsStr, ",")
--- a/internal/factsengine/gatherers/sudoers.go
+++ b/internal/factsengine/gatherers/sudoers.go
@@ -1,0 +1,302 @@
+package gatherers
+
+import (
+	"bufio"
+	"context"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/afero"
+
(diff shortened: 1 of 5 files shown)


CDimonaco (approved):
LGTM just some nitpicks

arbulu89 (commented):
Hey @balanza 
I would first ask Sanjeet if the output meets his expectations.
For example, just looking the output in the PR description, I find it difficult to understand.
What are the possible outputs? What are `commands_as_root`? Are there any other possible entries? Why do we have 3 different entries instead of one?

Besides that, I would try to flatten the output and have a list of commands. Maybe something like:
```
[
#{
command: "/usr/sbin/crm_attribute -n hana_prd_site_srHook_Site1 -v SOK -t crm_config -s SAPHanaSR",
no_password: true,
run_as_group: "",
run_as_user: "ALL"
},
#{
command: "/usr/sbin/crm_attribute -n hana_prd_site_srHook_Site1 -v SFAIL -t crm_config -s SAPHanaSR",
no_password: true,
run_as_group: "",
run_as_user: "ALL"
}
]
```

PD: For example, what is the output for this file coming in our [ref](https://documentation.suse.com/sbp/sap-15/html/SLES4SAP-hana-sr-guide-PerfOpt-15/index.html#id-allowing-sidadm-to-access-the-cluster):
```
# SAPHanaSR-ScaleUp entries for writing srHook cluster attribute
Cmnd_Alias SOK_SITEA    = /usr/sbin/crm_attribute -n hana_<sid>_site_srHook_<siteA> -v SOK   -t crm_config -s SAPHanaSR
Cmnd_Alias SFAIL_SITEA  = /usr/sbin/crm_attribute -n hana_<sid>_site_srHook_<siteA> -v SFAIL -t crm_config -s SAPHanaSR
Cmnd_Alias SOK_SITEB    = /usr/sbin/crm_attribute -n hana_<sid>_site_srHook_<siteB> -v SOK   -t crm_config -s SAPHanaSR
Cmnd_Alias SFAIL_SITEB  = /usr/sbin/crm_attribute -n hana_<sid>_site_srHook_<siteB> -v SFAIL -t crm_config -s SAPHanaSR
Cmnd_Alias HOOK_HELPER  = /usr/sbin/SAPHanaSR-hookHelper --sid=<SID> --case=checkTakeover

<sid>adm ALL=(ALL) NOPASSWD: SOK_SITEA, SFAIL_SITEA, SOK_SITEB, SFAIL_SITEB, HOOK_HELPER
```

PD: Would this gatherer find the entries in the file `/etc/sudoers.d/SAPHanaSR` with the previous content?

balanza:
@arbulu89 

>  I would first ask Sanjeet if the output meets his expectations.
For example, just looking at the output in the PR description, I find it difficult to understand.

The expectation is to check if the commands permissions for the SAP users are configured according to HA best practices. One of the checks is if the `/usr/sbin/SAPHanaSR-hookHelper` has the current configuration in the sudoers files.

> What are the possible outputs? What are commands_as_root? Are there any other possible entries? Why do we have 3 different entries instead of one?

This structure mimics the output of the `sudo` command in list mode. Commands can be a comma-separated list, thus the multiple entries in `commands_as_root`.

> I would try to flatten the output and have a list of commands

I like this actually, I will refactor accordingly thanks

internal/factsengine/gatherers/sudoers.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+type PrivilegeEntry struct {
+	RunAsUser  string   `json:"run_as_user"`
+	RunAsGroup string   `json:"run_as_group,omitempty"`
+	NoPassword bool     `json:"no_password"`
+	Commands   []string `json:"commands"`
+}
+
+type ParsedSudoers struct {
```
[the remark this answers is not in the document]
balanza: Done in https://github.com/trento-project/agent/pull/404/commits/f989a5ce3bd0ef594580a37ca2114b4595c5f503

internal/factsengine/gatherers/sudoers.go
```diff
@@ -0,0 +1,302 @@
...
+		Message: "error finding sudoers users",
+	}
+)
+
+type SudoersGatherer struct {
+	executor utils.CommandExecutor
+	fs       afero.Fs
+}
+
+type PrivilegeEntry struct {
```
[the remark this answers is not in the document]
balanza: Done in https://github.com/trento-project/agent/pull/404/commits/f989a5ce3bd0ef594580a37ca2114b4595c5f503

balanza:
@arbulu89 please check https://github.com/trento-project/agent/pull/404/commits/48cf880d6f48e23446b540557feefb196ed1852b, I refactor the output into a flatten structure

internal/factsengine/gatherers/sudoers.go
```diff
@@ -0,0 +1,298 @@
...
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// Check for the header line to start processing privilege entries.
+		if privilegesStartRegex.MatchString(line) {
+			inPrivilegesSection = true
+			continue // Skip the header line itself.
+		}
+
+		if inPrivilegesSection {
```
arbulu89, line 185: suggestion: I think we can use the early return technique to remove some indentation if we use it this way:
```
if !inPrivilegesSection {
    continue
}
```

internal/factsengine/gatherers/sudoers.go
```diff
@@ -0,0 +1,298 @@
...
+			commandsStr := matches[3]
+
+			var runAsUser, runAsGroup string
+			parts := strings.Split(runas, ":")
+			runAsUser = strings.TrimSpace(parts[0])
+			if len(parts) > 1 {
+				runAsGroup = strings.TrimSpace(parts[1])
+			}
+
+			noPassword := false
```
arbulu89, line 214: nitpick: `noPassword := flag == "NOPASSWD:"`

internal/factsengine/gatherers/sudoers.go
```diff
@@ -0,0 +1,298 @@
...
+
+func (g *SudoersGatherer) gatherAll(ctx context.Context) ([]parsedSudoers, *entities.FactGatheringError) {
+	usernames, err := findUsernames(g.fs)
+	if err != nil {
+		return nil, SudoersUserError.Wrap(err.Error())
+	}
+	allUsers := []parsedSudoers{}
+	for _, username := range usernames {
+		single, err := g.gatherSingle(ctx, username)
+		if err != nil {
```
arbulu89, line 122: comment: We need to be careful with this. It is a pretty common thing that SAP users might not remove the `/usr/sap/SID` directory once they remove the SAP installation. This would create "false" errors.
Maybe we should consider just continuing instead of failing and have a best-effort solution for the `gatherAll`. 
Later on, the check can implement to see if all real installations have their command.

arbulu89 (commented):
Really good job @balanza 
I left some comments. Specially the one where we can maybe decide to continue instead of fail in the `gatherAll`. We might have false errors, and maybe having a best-effort approach for that type is better

internal/factsengine/gatherers/sudoers.go
balanza: Uhm I didn't consider that, thanks for spotting. 

Let's go ahead and reason about it. Your point is fair, but we cannot just skip the error as it will inevitably lead to information hiding.

I think we can either:
1. fail the gathering (current code)
2. retrieve a list of facts which mix both data facts and error facts, and let the check inspect the result.

I'm not considering branching the behaviour based on the error type as it would require us to make too many assumptions IMHO. 

We have several aspects to consider:
1. We already have many gatherers that rely on the presence of the sid directory, thus it is something we already rely on as the _source of truth_
2. If a user does not clean up the filesystem after removing a SAP installation, there is no way for the checks to understand this. See the [SAP Profile gatherer](https://github.com/trento-project/agent/blob/a61a9b000d6b167ce1b20d907c63d2283b56611f/internal/factsengine/gatherers/sapprofiles.go#L57) for example.
3. Clean up _dangling_ directories after removing an installation should be considered a best practice and we should enforce it.

internal/factsengine/gatherers/sudoers.go
```diff
@@ -0,0 +1,298 @@
...
+func sidToUsername(sid string) string {
+	return strings.ToLower(sid) + "adm"
+}
+
+func findUsernames(fs afero.Fs) ([]string, error) {
+	usernames := []string{}
+
+	systemPaths, err := sapsystem.FindSystems(fs)
+	if err != nil {
+		return nil, SapProfilesFileSystemError.Wrap(err.Error())
```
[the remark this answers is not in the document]
balanza: Done in https://github.com/trento-project/agent/pull/404/commits/93276fb5b4ad192b6b45e51c8133da7e1f5d69a8

pkg/factsengine/entities/fact_value.go
```diff
@@ -58,6 +58,16 @@ func NewFactValue(factInterface interface{}, opts ...FactValueOption) (FactValue
 	}
 
 	switch value := factInterface.(type) {
+	case []string:
```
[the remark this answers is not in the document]
balanza: We cannot as then the item would be casted ad any as there is no intersection between string and interface, unfortunately.

internal/factsengine/gatherers/sudoers.go
balanza: Done in https://github.com/trento-project/agent/pull/404/commits/57164b0bd6af97fc2a76db1eba42a3721d246c7b

internal/factsengine/gatherers/sudoers.go
arbulu89: Well, but SAP profiles gatherer was specifically designed to check these kind of errors. We actually do have checks that test what you say.

In fact, in gatherers like `dispwork`, `sapcontrol` and `sapinstancehostnameresolver` we do skip the gathering of SAP instances without failing if one fails (and if I remember fine, we had this exact discussion about not failing in "empty" SAP installation folders, but I cannot find it).
At the end, this gatherer goal is not to find those issues.

> We already have many gatherers that rely on the presence of the sid directory, thus it is something we already rely on as the source of truth

Indeed, and most of them simply skip "invalid" SID entries.

> Clean up dangling directories after removing an installation should be considered a best practice and we should enforce it.

Yes, and we already have checks that fail if it is not done.

Now, up to you XD

pkg/factsengine/entities/fact_value.go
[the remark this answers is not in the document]
arbulu89: Can we at least add a test?

internal/factsengine/gatherers/sudoers.go
balanza: Done in https://github.com/trento-project/agent/pull/404/commits/4c4b72c9ddaced23a5a8947567505de6a0afafaa

arbulu89: Thank you!
Really appreciated. :green_circle: light from my side

PR #407, opened 2025-03-05
Update permissions of agent.yaml

janvhs:
Only the root user is allowed to read the file now


abravosuse (approved):
LGTM @janvhs

EMaksy (approved):
LGTM, just curious what was the reason to change this ? :D
To hide the api key ?

PR #410, opened 2025-03-13
Fix empty element parsing

nelsonkopliku:
# Description

Reported [bug](https://bugzilla.suse.com/show_bug.cgi?id=1239534) mentions that aws metadata discovery panics because of a `runtime error: slice bounds out of range [-1:]`

```
> trento-agent[26856]: time="2025-03-06 11:07:04" level=info msg="System is running on aws"
> trento-agent[26856]: panic: runtime error: slice bounds out of range [-1:]
> trento-agent[26856]: goroutine 59 [running]:
> trento-agent[26856]: github.com/trento-project/agent/internal/core/cloud.buildAWSMetadata(...)
> trento-agent[26856]: #011github.com/trento-project/agent/internal/core/cloud/aws.go:124
```

I could not reproduce against an AWS vm of ours, but what the error says is that we are attempting the following
```go
if element[len(element)-1:] == "/" {
```
on an empty slice (an empty string in this case).

If the fix makes sense we might provide the customer with a patched agent to test.


test/fixtures/discovery/aws/broken
```diff
@@ -0,0 +1,3 @@
+foo
+bar
+next-line-must-be-empty
```
nelsonkopliku, line 3: I swear there is an empty line 4 😅

arbulu89 (approved):
@nelsonkopliku I think we can merge this as well. Together with the imds v2 change.
it makes the code more resillient

nelsonkopliku:
Closing this because it ended up in https://github.com/trento-project/agent/pull/414

@gagandeepb we're using `strings.TrimSpace`. It should do

PR #414, opened 2025-03-24
Support AWS IMDSv2

nelsonkopliku:
# Description

This PR aims to add support for IMDSv2, currently trento-agent assumes IMDSv1 always enabled.

With this change the agent is able to still discover AWS as a provider, the worst thing that can happen is that it cannot extract metadata.

Refs:
- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
- https://aws.amazon.com/blogs/aws/amazon-ec2-instance-metadata-service-imdsv2-by-default/

Note that current change is targeting a branch detached from the last 2.4.0 release, so that if we need to release a PTF we have only the necessary changes.


internal/core/cloud/metadata_test.go
```diff
@@ -302,6 +311,31 @@ func (suite *CloudMetadataTestSuite) TestNewCloudInstanceAWS() {
 	suite.Equal("some-id", meta.InstanceID)
 }
 
+func (suite *CloudMetadataTestSuite) TestNewAWSCloudInstanceWithoutMetadata() {
+	ctx := context.TODO()
+	suite.mockExecutor.
+		On("Exec", "dmidecode", "-s", "chassis-asset-tag").
+		Return(dmidecodeEmpty(), nil).
+		On("Exec", "dmidecode", "-s", "system-version").
+		Return(dmidecodeAWSSystem(), nil)
```
nelsonkopliku, line 320: In order to improve decoupling and testability provider detection needs to be refactored, imho, by possibly introducing an interface.

However this would be out of scope here and provider detection is going to be evolved anyways when we address PPC support.
[A gist on what I mean here](https://github.com/trento-project/agent/pull/413/files#diff-f858dae8a26664ac95b63cfea1249fbcb07edec4be48e4d7f064d2f839d5a4a1R20)

arbulu89 (commented):
Hey @nelsonkopliku 
I started doing the review, but before going with details, i have some few questions:

With the current code, every discovery loop we create a new `AWSMetadata` object, which at the beginning, it requests the token. This means that we are getting a new token every 10 seconds (by default).

Based on this. Why do we need to check for token authorization every query? I mean, it is certain that it won't be expired. The flow would be simple:

1. get token
2. fails. We just finish with error. continue otherwise
3. we do the query loop. with 6 hours ttl (1 minute as well), we surely know that we will be authorized

I mean, it looks like the code was prepared to reuse the token in subsequent discovery loops, but it is not. 
If we reuse the token on them, yes, we need to do more handling but I don't see this happening. Or am I missing something?

nelsonkopliku:
Hey @arbulu89, is this what you're looking for?
https://github.com/trento-project/agent/pull/414/files#diff-960d16329d4bc12e6b757326436131b46688359b53ac281ab36c23affe43b1e3R268-R271

```go
if metadataFetchContext.HasToken() {
	log.Debug("Using previously fetched IMDS token")
	return metadataFetchContext, nil
}
```
In `requestMetadataToken` if `metadataFetchContext` has already a token it will not query for a new one.

Or am _ _I_ _  missing something 😅 ?

nelsonkopliku:
Hey @arbulu89 at the end I went for what currently is simpler: we get a new token on every discovery loop.
I realized that it would be just one extra request to IMDS. We already to dozens at every tick, and one more is not gonna be a big issue 🙈 

---

For reference here https://github.com/trento-project/agent/compare/detached-release-2.4.0...imds-v2?expand=1 we have an alternative approach that makes sure to reuse the token among discovery loops and refresh when needed, in case we want to reconsider in the future.
The best learning from that attempt was:
- an extra confirmation that a state in the agent might be helpful
- if we want to use an in-memory cache as a potential state (even though transient among agent restarts) we might want to pull a library doing that

internal/core/cloud/aws.go
```diff
@@ -151,11 +173,45 @@ func requestMetadata(ctx context.Context, client HTTPClient, url string) (interf
...
 		err := json.Unmarshal(body, &jsonData)
 		return jsonData, err
 	}
 	return string(body), nil
 }
+
+func requestMetadataToken(ctx context.Context, client HTTPClient) (string, error) {
+	log.Debug("Fetching IMDS token...")
+
+	url := fmt.Sprintf("%sapi/token", awsMetadataURL)
```
arbulu89, line 192: nitpick: we could most probably use [path.Join](https://pkg.go.dev/path#Join). I know that we use `Sprintf` in many places though :sweat_smile:

internal/core/cloud/aws.go
```diff
@@ -151,11 +173,45 @@ func requestMetadata(ctx context.Context, client HTTPClient, url string) (interf
...
+	if err != nil {
+		log.Debugf("An error occurred while fetching IMDS token: %s", err)
+		return "", err
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.Errorf("failed to fetch metadata token: %s", resp.Status)
+	}
+
+	defer resp.Body.Close()
```
arbulu89, line 208: nitpick: should be maybe run this `defer` before the previous `if` conditions?

internal/core/cloud/aws.go
```diff
@@ -151,11 +173,45 @@ func requestMetadata(ctx context.Context, client HTTPClient, url string) (interf
...
 	if json.Valid(body) {
-		var jsonData interface{}
+		var jsonData any
 		err := json.Unmarshal(body, &jsonData)
 		return jsonData, err
 	}
 	return string(body), nil
 }
+
+func requestMetadataToken(ctx context.Context, client HTTPClient) (string, error) {
```
arbulu89, line 189: super nitpick: as this is the first function we use in `NewAWSMetadata`, could we put it just after it? to avoid that much scrolling

internal/core/cloud/aws.go
```diff
@@ -12,12 +12,22 @@ import (
...
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 )
 
 const (
-	awsMetadataURL      = "http://169.254.169.254/latest/"
-	awsMetadataResource = "meta-data"
+	awsMetadataURL                = "http://169.254.169.254/latest/"
+	awsMetadataResource           = "meta-data"
+	metadataTokenTTLHeader string = "X-aws-ec2-metadata-token-ttl-seconds" //nolint
```
arbulu89, line 25: question: why do we need these `nolint`?

internal/discovery/collector/publishing_test.go
```diff
@@ -4,20 +4,23 @@ package collector_test
...
 	"github.com/trento-project/agent/internal/discovery/collector"
 	"github.com/trento-project/agent/internal/discovery/mocks"
 	"github.com/trento-project/agent/test/helpers"
 )
 
 const (
-	apiKey        = "some-api-key"
-	discoveryType = "sap_system_discovery"
+	apiKey             = "some-api-key"
+	sapSystemdiscovery = "sap_system_discovery"
```
arbulu89, line 22: nitpick: most probably we should remove this `global` const and have the discovery type in each test. What do you think?

internal/discovery/collector/publishing_test.go
```diff
@@ -85,27 +88,79 @@ func (suite *PublishingTestSuite) TestCollectorClientPublishingSubscriptionDisco
...
+			},
+		},
+	}
+
+	suite.runDiscoveryScenario(cloudDiscovery, discoveredCloudInstance, func(requestBodyAgainstCollector string) {
+		suite.assertJSONMatchesJSONFileContent(helpers.GetFixturePath("discovery/aws/collector_payloads/expected_published_cloud_discovery_with_metadata.json"), requestBodyAgainstCollector)
+	})
+}
+
+func (suite *PublishingTestSuite) TestCollectorClientPublishingAWSCloudDiscoveryWithoutMetadata() {
```
arbulu89, line 132: nitpick: maybe you could include this test with the final one. At the end, they look the same, don't they?

arbulu89 (approved):
Great @nelsonkopliku !
I do agree that we don't really need premature optimization by now. Asking one token per discovery is enough (most probably, because maybe we don't even need to query the discovery every 10 seconds, but that's discussion for other time hehe).

2 final questions:
- Do we still fix the error of trying to access a 0 length array? Do we have other PR for this? I think it is worthy to fix it
- Now I guess, we need to test web, to see if sending an empty metadata for AWS needs some adjustments

nelsonkopliku:
> * Do we still fix the error of trying to access a 0 length array? Do we have other PR for this? I think it is worthy to fix it
>
I might bring that here from the other PR https://github.com/trento-project/agent/pull/410
The point is what we want to release how: meaning what goes in a PTF, if we want any or if we wait these changes to go out with next release

> * Now I guess, we need to test web, to see if sending an empty metadata for AWS needs some adjustments
>
I have this in draft https://github.com/trento-project/web/pull/3412

internal/core/cloud/aws.go
nelsonkopliku: Tried it, however besides looking more related to filesystem paths, it breaks with the `http://...`

```go
url := path.Join(awsMetadataURL, "api/token")
```

returns `"http:/169.254.169.254/latest/api/token"`
Note that there's one missing `/` after `http:`

internal/core/cloud/aws.go
nelsonkopliku: Shouldn't be necessary, should it? 
I mean that if we return in previous if conditions, the body is not read at all, so we don't need defering closing it. Am I missing something?

internal/core/cloud/aws.go
nelsonkopliku: Here it is 3cdd69d4a12a1305a7f8949908db31a354781128

internal/core/cloud/aws.go
nelsonkopliku: golang ci lint complains 🙈 

```
internal/core/cloud/aws.go:25:2: G101: Potential hardcoded credentials (gosec)
        metadataTokenTTLHeader string = "X-aws-ec2-metadata-token-ttl-seconds"
        ^
internal/core/cloud/aws.go:26:2: G101: Potential hardcoded credentials (gosec)
        metadataTokenHeader    string = "X-aws-ec2-metadata-token"
```

internal/discovery/collector/publishing_test.go
nelsonkopliku: I don't mind either way. There's just some extra repetition but I'm fine with it.
Here's the commit a5bddc2b5eeb66d80faa8375c32f73406cc410e1

internal/core/cloud/aws.go
arbulu89: I always thought that we needed to close it if we get a not errored response from the client queries, but I haven't checked more than that

arbulu89: I found this for example: https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body
```
If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.
```
I would at least use the defer before doing the forbidden/authorized check

internal/discovery/collector/publishing_test.go
nelsonkopliku: Lol there was a reason, ci lint fails.
Fixed here 8273b25694c04d26122a9c43e33e8764cf308544

internal/discovery/collector/publishing_test.go
nelsonkopliku: Alright a190a0ac60953af98fa5cd562f9be68202ddb31f

gagandeepb (commented):
Perhaps not relevant for this PR, but what do you think about pulling in the AWS go sdk? A part of it is auto-generated (IIRC) from the latest AWS service definitions and should give us the intended EC2 metadata out-of-the-box.

nelsonkopliku:
> Perhaps not relevant for this PR, but what do you think about pulling in the AWS go sdk? A part of it is auto-generated (IIRC) from the latest AWS service definitions and should give us the intended EC2 metadata out-of-the-box.

That's a good point @gagandeepb. Definitely worth exploring as an evolution path for the discovery process.

balanza:
> Perhaps not relevant for this PR, but what do you think about pulling in the AWS go sdk? A part of it is auto-generated (IIRC) from the latest AWS service definitions and should give us the intended EC2 metadata out-of-the-box.

Plus one. We may get rid of a lot of custom code and have a solid solution (aws team knows better than us the integration pitfalls that may occur)

arbulu89:
Hey @nelsonkopliku ,
Don't forget to bring the change to `main` as well at some point.

nelsonkopliku:
> Don't forget to bring the change to `main` as well at some point.

Sure, here it is https://github.com/trento-project/agent/pull/423

PR #418, opened 2025-03-27
Identify current instance SAP instances

arbulu89:
# Description

When we query the current installed SAP system instances using `sapcontrol`, we cannot be 100% of which of them is the one currently running if the virtual hostname are duplicated.
Until now we did this identification in web: https://github.com/trento-project/web/blob/main/lib/trento/discovery/payloads/sap_system_discovery_payload.ex#L269

It is based on checking the instance number and saplocalhost values from the current profile with the values in the sapcontrol instances list. But if the hostnames are duplicated, we were simply returning the first catch.

A customer came with a scenario like that, where 2 ERS instances have the same virtual hostname. 

Now, we will check this in the agent using some local files that SAP creates during one of its binaries execution, where the content says clearly in which real host this SAP instance is running.

Here some references about the implementation: https://m1bc.home.blog/2019/09/09/getsysteminstancelist-duplicate-entries/

PD: I'm still not 100% sure about the `runningLocally` field name. This field is used to determine if the entry in the sapcontrol instances list is the currently discovered instance. Maybe `runningLocally` is missleading, because the instance could be running locally in this machine but not be the current instance. Maybe I will rename to `currentInstance` or something else

## How was this tested?
UT and manual testing

### DIFF
--- a/internal/core/sapsystem/sapcontrol.go
+++ b/internal/core/sapsystem/sapcontrol.go
@@ -40,8 +40,8 @@ func NewSAPControl(ctx context.Context, w sapcontrol.WebService, fs afero.Fs, ho
 		Instances:  instances.Instances,
 	}
 
-	if err := sapControl.enrichRunningLocally(fs, hostname); err != nil {
-		return nil, errors.Wrap(err, "Error finding locally running instances")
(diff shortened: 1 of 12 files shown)


CDimonaco (approved):

nelsonkopliku (approved):
LGTM

PR #420, opened 2025-04-08
Generic event handler and messaging

arbulu89:
# Description

Some preliminary work to enable the addition of operations. I didn't want to bring the operations code all together, it looked to big.
The PR basically:
- Move the messaging to a new package
- Creates a new event handler utility, that it will be used by the facts engine and operations
- Updates some dependencies including workbench for example
- Make the `policy` code more functional and public to test better

## How was this tested?
UT


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/trento-project/agent
 
-go 1.23
+go 1.23.4
+
+toolchain go1.23.6
```
arbulu89, line 5: I don't honestly know why the `go get` and `go tidy` are putting this. We can remove them manually though

balanza (approved):
LGTM

go.mod
balanza: See no evil

PR #421, opened 2025-04-08
Use absolute command paths

balanza:
Use an absolute path when executing an external program on the host machine. 

Paths are defined according to a standard installation of `SUSE Linux Enterprise Server 15`

### DIFF
--- a/internal/factsengine/gatherers/ascsers_cluster_test.go
+++ b/internal/factsengine/gatherers/ascsers_cluster_test.go
@@ -60,7 +60,7 @@ func (suite *AscsErsClusterTestSuite) TestAscsErsClusterGatherCmdNotFound() {
 
 func (suite *AscsErsClusterTestSuite) TestAscsErsClusterGatherCacheCastingError() {
 	cache := factscache.NewFactsCache()
-	_, err := cache.GetOrUpdate("/usr/sbin/cibadmin", func(_ ...interface{}) (interface{}, error) {
+	_, err := cache.GetOrUpdate("cibadmin", func(_ ...interface{}) (interface{}, error) {
 		return 1, nil
 	})
 	suite.NoError(err)
@@ -280,7 +280,7 @@ func (suite *AscsErsClusterTestSuite) TestAscsErsClusterGather() {
 
 	entries := suite.cache.Entries()
 	expectedEntries := []string{
-		"/usr/sbin/cibadmin",
+		"cibadmin",
 		"sapcontrol:GetProcessList:PRD:00",
 		"sapcontrol:GetProcessList:PRD:10",
 		"sapcontrol:GetProcessList:DEV:01",
--- a/internal/factsengine/gatherers/ascsers_cluster_test.go
+++ b/internal/factsengine/gatherers/ascsers_cluster_test.go
@@ -38,7 +38,7 @@ func (suite *AscsErsClusterTestSuite) SetupTest() {
 }
 
 func (suite *AscsErsClusterTestSuite) TestAscsErsClusterGatherCmdNotFound() {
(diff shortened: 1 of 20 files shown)


gagandeepb (commented):
I am not fully aware of the motivation behind this change. I wonder what are the underlying system scenarios under which the agent execution thread is not able to find the `dmidecode` installed system package from it's default installation location? What makes this change necessary now/at this point in time, as compared to previous versions of the agent and its execution/system environment ?

internal/factsengine/gatherers/ascsers_cluster_test.go
```diff
@@ -280,7 +280,7 @@ func (suite *AscsErsClusterTestSuite) TestAscsErsClusterGather() {
 
 	entries := suite.cache.Entries()
 	expectedEntries := []string{
-		"cibadmin",
+		"/usr/sbin/cibadmin",
```
arbulu89, line 283: Should this be here?
The cache only needs to know that `cibadmin` output was stored, not the full path most probably

internal/factsengine/gatherers/cibadmin.go
```diff
@@ -10,8 +10,8 @@ import (
 )
 
 const (
-	CibAdminGathererName  = "cibadmin"
-	CibAdminGathererCache = "cibadmin"
+	CibAdminGathererName  = "/usr/sbin/cibadmin"
```
arbulu89, line 13: Careful!
Don't touch this. This references the name of the gatherer and the name of the entry in the cache

arbulu89 (commented):
I think you changed accidentally the name of the `cibadmin` gatherer and the cache name.

I have to confess that this changes makes me a bit anxious, as a change on the folder placement of any of the binaries on some SP pack would break things. And of course we cannot test this with mocked agents.

Either way, I guess that this change was somehow agreed (I was not aware of current efforts to make the agent run without being root).

balanza:
@janvhs, thanks for the long explanation, it's very helpful indeed.

Changes have been made looking at the current supported platforms (SLES4SAP 15 SP3 and above). It might be fine for standard installations; nevertheless, I agree it's not robust nor future-proof.

> [...] implement a PATH-independent lookup mechanism

This might be a better solution, I'll look at it.

-----

> I guess that this change was somehow agreed (I was not aware of current efforts to make the agent run without being root)

@arbulu89, we didn't discuss it deeply. We have some security concerns derived from the fact that the agent runs with root privileges. This PR start polishing the agent so that we can refine the required capabilities and run under an unprivileged user.

gagandeepb (commented):
Few comments/questions:

1. What are your thoughts about making the full installation paths of runtime external dependencies like `dmidecode` configurable at runtime via e.g. a config file, with some implied defaults if not configured? This would reduce coupling to specific installation locations/paths which might change. It would also reduce repetition in the code base.

2. Additionally, do we have any evidence/support for whether dmidecode/similar external deps (e.g. on SLES4SAP) are meaningfully invokable (e.g. provides the necessary output) without the calling process being run as root itself ?

internal/factsengine/gatherers/ascsers_cluster_test.go
balanza: That's a nasty typo, I'm glad you found it. Fixed in https://github.com/trento-project/agent/pull/421/commits/14909a1fcde6eb56f24e95d4298db38f2537f1b5

internal/factsengine/gatherers/cibadmin.go
balanza: Fixed in https://github.com/trento-project/agent/pull/421/commits/14909a1fcde6eb56f24e95d4298db38f2537f1b5

arbulu89 (approved):
I'm fine.
With this change the need to test different service packs more often arises, specially when SLE16 closer than ever.

PR #422, opened 2025-04-08
Run operations

arbulu89:
# Description

Listen for operation requests and run them with `workbeanch`.

Nothing complex, it more or less replicates what we do with the facts gathering

## How was this tested?

UT

### DIFF
--- a/internal/operations/policy.go
+++ b/internal/operations/policy.go
@@ -1,0 +1,75 @@
+package operations
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/trento-project/agent/internal/messaging"
+
+	"github.com/trento-project/contracts/go/pkg/events"
+	"github.com/trento-project/workbench/pkg/operator"
+)
+
+const (
+	OperatorExecutionRequestedV1 = "Trento.Operations.V1.OperatorExecutionRequested"
+)
+
+func HandleEvent(
+	ctx context.Context,
+	event []byte,
+	agentID string,
+	adapter messaging.Adapter,
+	registry operator.Registry,
+) error {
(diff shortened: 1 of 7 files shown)


internal/operations/policy.go
```diff
@@ -0,0 +1,75 @@
...
+
+		completedEvent, err := OperatorExecutionCompletedToEvent(
+			operatorExecutionRequested.OperationID,
+			operatorExecutionRequested.GroupID,
+			target.AgentID,
+			operatorExecutionRequested.StepNumber,
+			report,
+		)
+		if err != nil {
+			return err
```
balanza, line 59: **issue:** ‏Let's wrap this error, too.


**question:** ‏Returning an error in this case discards the event from the queue, which can re-emit the same event (according to the queue policies). As this error comes after the actual operation execution, is it safe to do so?

internal/operations/policy.go
```diff
@@ -0,0 +1,75 @@
...
+
+		target := operatorExecutionRequested.GetTargetAgent(agentID)
+		if target == nil {
+			log.Infof("OperatorExecutionRequested is not for this agent. Discarding operator execution")
+			return nil
+		}
+
+		operatorBuilder, err := registry.GetOperatorBuilder(operatorExecutionRequested.Operator)
+		if err != nil {
+			return err
```
balanza, line 46: **issue:** ‏ It's better to wrap this error as it's coming from an external codebase, and it might not be straightforward to troubleshoot in case of failure.

arbulu89: Done!

internal/operations/policy.go
arbulu89: Error message wrapped

balanza (approved):

CDimonaco (approved):

PR #424, opened 2025-04-14
Listen discovery requests

arbulu89:
# Description

Listen to discovery requests.
It handlers the `DiscoveryRequested` event in a new rabbitmq queue and runs the requested discovery.
All existing discoveries are allowed. I don't find any reason to whitelist things by now.

Related to:
https://github.com/trento-project/web/pull/3443
https://github.com/trento-project/contracts/pull/98

## How was this tested?

UT and manual testing


nelsonkopliku (approved):
@arbulu89 in general looks good to me.

However, I am thinking about scenarios where ongoing discovery ticks run concurrently with an on demand request for the same discovery type.

I am not much concerned about two discoveries being published back to back, the last publish will possibly override the first one.

What I am wondering is if we should be careful about _heavy_ discoveries running concurrently.

Not that we should do anything here and now, just asking.

arbulu89:
@nelsonkopliku yes, that's a natural concern, I have it myself hehe
And not that much about discoveries, which are just read only and light on nature. I'm more concerned about what would happen when a discovery and operation happen concurrently. Some commands and actions we do in them might have their own concurrency problems.
Right now, the number of operations and discovery request is minimal so I don't think we will have issues, but it is something we should check future wise.

I didn't want to add any premature optimization by now, but I agree, something to have in our radar. Maybe some sort of locking mechanism for concurrent discoveries/operations. Something to be discussed among developers and with @abravosuse

PR #428, opened 2025-05-06
Streamline CI workflow

stefanotorresi:
# Description

- Removes the "Cutting Edge" rolling release from GitHub actions, and rely on OBS factory repos instead.
- Removes the artifacts publishing job, since we don't test those artifacts anywhere and they are intended to be replaced by continuously built rpms in OBS.
- Removes the OBS automatic submit, since it was quirky and often source of confusion (this step was only in place for the Agent and not in any other package, oddly enough).
- Removes the tgz direct download installation method from the simplified installation script
- Removes the `gh_release_to_obs_changeset.py` hack script, since it has been moved to the [trento/continuous-delivery](https://github.com/trento-project/continuous-delivery) repo a while ago.


install-agent.sh
```diff
@@ -144,21 +136,16 @@ function install_trento() {
...
 
 function install_trento_rpm() {
     if [[ -n "$USE_ROLLING" ]]; then
-        TRENTO_REPO=${TRENTO_REPO:-"https://download.opensuse.org/repositories/devel:/sap:/trento:/factory/15.3/devel:sap:trento:factory.repo"}
-        TRENTO_REPO_KEY=${TRENTO_REPO_KEY:-"https://download.opensuse.org/repositories/devel:/sap:/trento:/factory/15.3/repodata/repomd.xml.key"}
+        TRENTO_REPO=${TRENTO_REPO:-"https://download.opensuse.org/repositories/devel:/sap:/trento:/factory/standard/devel:sap:trento:factory.repo"}
+        TRENTO_REPO_KEY=${TRENTO_REPO_KEY:-"https://download.opensuse.org/repositories/devel:/sap:/trento:/factory/standard/repodata/repomd.xml.key"}
     else
-        TRENTO_REPO=${TRENTO_REPO:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/devel:sap:trento.repo"}
-        TRENTO_REPO_KEY=${TRENTO_REPO_KEY:-"https://download.opensuse.org/repositories/devel:/sap:/trento/15.3/repodata/repomd.xml.key"}
```
stefanotorresi, line 161: Note the the reference here is no longer hard coded to 15.3, which was quite an oversight.
The OBS project now provides a "standard" repo, which is an alias of the latest SP (currently SP6).
See https://build.opensuse.org/projects/devel:sap:trento:factory/meta for more details.

nelsonkopliku (commented):
With this change basically there won't be any artifact in github anymore, not even for an actual release.

I am referring to the Assets section like here https://github.com/trento-project/agent/releases/tag/2.4.0
image

stefanotorresi:
> With this change basically there won't be any artifact in github anymore, not even for an actual release.
> 
> I am referring to the Assets section like here https://github.com/trento-project/agent/releases/tag/2.4.0 image

Yes, that is intended. These artifacts don't have an actual use case, and are not tested in any way.

PR #439, opened 2025-05-27
update installation script and readme instructions

stefanotorresi:
# Description

- Removes the reference to the current version number in the script (this is required for the new release workflow).
- Follow ups to #428 in further reducing the code footprint of the installation script by removing unused code and changing how the systemd unit was handled.
- Updates the repo README.md to make `zypper in` the most prominent and authoritative installation option, while also making the shell script the least prominent one.

## Did you add the right label?

- [x] **DONE**

## How was this tested?

Manual testing on openSUSE 15.5 vm.

- [x] **DONE**

## Did you update the documentation?

- [x] **DONE**


gagandeepb (commented):
I wonder if you think agent <> web/wanda version compatibility/version drift is a concern here. For example, should this script be more opinionated/prescriptive in suggesting/doing these version checks?

gagandeepb (approved):

stefanotorresi:
>I wonder if you think agent <> web/wanda version compatibility/version drift is a concern here. For example, should this script be more opinionated/prescriptive in suggesting/doing these version checks?

Nah, not at all, the script is simplistic and super barebone, it doesn't even handle the difference between openSUSE and SLES atm.
To be honest, it's only here for vestigial reasons, it's not even enterprise-supported, and I will be happier the day we remove it entirely.

Component version drift is currently handled at the packaging level: as long as we ensure that the helm chart and the rpm packages are correctly expressing interdependencies, we should be fine.

PR #444, opened 2025-06-03
Disable Power PC Arch

janvhs:
# Description

As discussed, we don't want to support the agent on PPC64


stefanotorresi:
Correct, we need to remove aarch64, not ppc64le.

PR #449, opened 2025-06-09
Replace logrus logger with built-in log/slogs

balanza:


gagandeepb (commented):
I wonder what you think about renaming the slog import as `log` (instead of the default `slog`) to reduce the diff here in this PR ?

gagandeepb (approved):

PR #455, opened 2025-07-01
Systemd units discovery

nelsonkopliku:
# Description

This PR enhances the host discovery by adding relevant information about relevant systemd units.

As of now it only returns whether the unit file is enabled/disabled at boot and only `pacemaker.service` is among the relevant units.

I considered the alternative to return a list like
```json
"enabled_services":["foo.service", "bar.service"]
```
but that turns out to be quite limited and quite soon because:
- we might need some other information besides the file status
- not being in `enabled_services` list would not necessarily mean being disabled - ie we might not have been able to determine whether it is enabled or not

That said the proposed shape turns out being more flexible and future-proof.

Another consideration: I couldn't find any function that allows us to retrieve the needed information for a list of units [see doc](https://pkg.go.dev/github.com/coreos/go-systemd/v22/dbus), that's why we do it in a loop.
I didn't want to add concurrent systemd calls for a list of 1 element now 😅 If the list of the relevant units grows substantially we can consider doing the extraction concurrently.

Next:
- handle in web in a backward compatible manner (?)
- show the information somewhere in the ui
- use the new information to allow/disallow relevant operations

## How was this tested?

Automatic tests.

### DIFF
--- a/internal/core/hosts/host.go
+++ b/internal/core/hosts/host.go
@@ -8,42 +8,35 @@ import (
 )
 
 type UnitInfo struct {
+	Name          string `json:"name"`
 	UnitFileState string `json:"unit_file_state"`
 }
 
-type SystemdUnitsStatus map[string]UnitInfo
-
-func DefaultUnitInfo(units ...string) SystemdUnitsStatus {
-	defaultInfo := make(SystemdUnitsStatus)
-	if len(units) == 0 {
-		units = []string{"pacemaker.service"}
-	}
+func defaultUnitInfo(units []string) []UnitInfo {
+	defaultInfo := []UnitInfo{}
 	for _, unit := range units {
-		defaultInfo[unit] = UnitInfo{
+		defaultInfo = append(defaultInfo, UnitInfo{
+			Name:          unit,
 			UnitFileState: "unknown",
-		}
+		})
 	}
 	return defaultInfo
 }
 
-func GetSystemdUnitsStatus(ctx context.Context, units ...string) SystemdUnitsStatus {
+func GetSystemdUnitsStatus(ctx context.Context, units []string) []UnitInfo {
 	dbus, err := systemd.NewDbusConnector(ctx)
 	if err != nil {
 		slog.Error("Error while creating dbus connection", "error", err)
 	}
-	return getSystemdUnitsStatus(ctx, dbus, units...)
+	return GetSystemdUnitsStatusWithCustomDbus(ctx, dbus, units)
 }
 
 func GetSystemdUnitsStatusWithCustomDbus(
 	ctx context.Context,
 	dbus systemd.DbusConnector,
-	units ...string,
-) SystemdUnitsStatus {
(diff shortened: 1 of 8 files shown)


internal/core/hosts/host.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	return defaultInfo
+}
+
+func GetSystemdUnitsStatus(ctx context.Context, units ...string) SystemdUnitsStatus {
+	dbus, err := systemd.NewDbusConnector(ctx)
+	if err != nil {
+		slog.Error("Error while creating dbus connection", "error", err)
+	}
+	return getSystemdUnitsStatus(ctx, dbus, units...)
```
arbulu89, line 34: nitpick: maybe we could have `return GetSystemdUnitsStatusWithCustomDbus(...` so it gives more sense of coverage, as the 2nd is explicitly tested

internal/core/hosts/host.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/trento-project/agent/internal/core/hosts/systemd"
+)
+
+type UnitInfo struct {
+	UnitFileState string `json:"unit_file_state"`
+}
+
+type SystemdUnitsStatus map[string]UnitInfo
+
+func DefaultUnitInfo(units ...string) SystemdUnitsStatus {
```
arbulu89, line 16: I find this `DefaultUnitInfo` usage a bit oriented to have just more individual testing.
It doesn't really look like something we need as a public function.

Just my opinion, but my approach would be:

In `discovery/host.go` pass the services we want to discovery, without default values:
```
const systemdUnits = []string{"pacemaker.service"}
...
SystemdUnits:  hosts.GetSystemdUnitsStatus(ctx, discoveryUnits),
```

And here in the loop of `getSystemdUnitsStatus`:
```
defaultInfo := make(SystemdUnitsStatus)
...
# or the flatten version of it
for unit := range units {
    defaultInfo[unit] = UnitInfo{
        UnitFileState: "unknown",
     }

     ...
```

arbulu89 (approved):
Looks good!
I think sending the unit files info as struct is a good idea.
In the other hand, I would vote more for a flatten version of the values:
```
"systemd_units": [
 {
   "unit": "pacemaker.service",
    "unit_file_state": "enabled"
}
] 
```

Generally it provides a more flexible way of using the data with our typical functional programming methods. 
I understand that now the easy access of the map looks better, as we only want to expose one service and in one specific place. But if we expose more, and for example, we want to expose them in the host details view, the flatten version usually comes more handy.

I have noted some additional code-wise comments, but feel free to decide what to do.

nelsonkopliku:
Hey @arbulu89 thanks for the feedback, it was exactly what I was expecting 😄 

Here we go ced05f965cb76e3a23770e53a37ef3bcd87a4dad

arbulu89 (approved):
:green_heart: 
Really nice!

PR #475, opened 2025-08-18
Improve agent docs for auto build

EMaksy:
# Description

Just a small update of the docs structure for the auto build pr


nelsonkopliku (approved):
👀

PR #578, opened 2026-04-30
Add status gatherer

balanza:
A simple gatherer that is meant to signal that the agent is running and able to accept data gathering requests. It is meant for testing the checks infrastructure to ensure it is working properly.

### DIFF
--- a/cmd/facts.go
+++ b/cmd/facts.go
@@ -11,12 +11,14 @@ import (
 	"os/signal"
 	"syscall"
 
+	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
 	"github.com/trento-project/agent/internal/agent"
 	"github.com/trento-project/agent/internal/factsengine/gatherers"
+	"github.com/trento-project/agent/internal/identity"
 	"github.com/trento-project/agent/pkg/factsengine/entities"
 	"github.com/trento-project/agent/pkg/utils"
 )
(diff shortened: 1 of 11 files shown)


internal/factsengine/gatherers/gatherer.go
```diff
@@ -23,6 +23,9 @@ type FactGathererWithCache interface {
 
 func StandardGatherers() FactGatherersTree {
 	return FactGatherersTree{
+		StatusGathererName: map[string]FactGatherer{
```
arbulu89: nit:
let's put in its corresponding alphabetical order

internal/factsengine/gatherers/status.go
```diff
@@ -0,0 +1,77 @@
...
+	slog.Info("Starting facts gathering process", "gatherer", StatusGathererName)
+
+	machineIDBytes, err := afero.ReadFile(g.fs, g.machineIDPath)
+	if err != nil {
+		slog.Error("Error reading machine ID", "error", err)
+		return facts, StatusMachineIDError.Wrap(err.Error())
+	}
+
+	machineID := strings.TrimSpace(string(machineIDBytes))
+	agentID := uuid.NewSHA1(trentoAgentNamespace, []byte(machineID)).String()
```
arbulu89: question:
cannot we use `agent.GetAgentID` function here?
at the end, it does the same.
i don't see we get much versatility having a variable machine id path. At the end, we can mock it the same way.
the unique inconvenience I can think of is something cyclical import issue.

arbulu89 (approved):
Looks good!
just a comment about using an already existing function to get the agent ID.
Anyway, i like the idea of having this for our tests.

internal/factsengine/gatherers/status.go
balanza: I moved the function into a dedicated module.

Moreover, to handle the `force-agent-id` config, I had to pass down the configuration to the gatherer.

cmd/facts.go
```diff
@@ -90,7 +92,15 @@ func gather(cmd *cobra.Command, _ []string) {
 
 	slog.SetDefault(logger)
 
-	gathererRegistry := gatherers.NewRegistry(gatherers.StandardGatherers())
+	agentID, err := identity.GetAgentID(afero.NewOsFs())
+	if err != nil {
+		slog.Error("Could not derive agent ID", "error", err)
```
arbulu89, line 97: nit:
could we maybe print: `slog.Errorf("could not get the agent ID", "error", err)` to have the same message than in other commands

arbulu89 (approved):
Thank you!
Maybe run some manual tests if you didn't 🙂 
It looks a pretty sensitive place

PR #623, opened 2026-07-20
Add error collect payload to log msg

antgamdia:
### Description

This PR enriches the error message when the Collector client fails to push the data to `trento-web`.  Now, when publishing data to the collector fails, the error message will include the response body, making debugging easier. 

Related TRNT-4563

### How was this tested?

IRL test

### Documentation changes

No

### Additional information

N/A


nelsonkopliku (approved):
I am just a bit concerned about the fact that the output might be heavier and so the footprint on filesystem (when logs are stored in the filesystem).

The agent makes around 70 collection requests per minute, however since it's only for errored collections, we hopefully will not flood the logs with too much extra content (assuming the failures are not that frequent 😅) and I see it might be helpful to troubleshoot.

In any case this makes me think about the following:

Considering that kind-of-same info can be found in web's logs, I am keen to think we should invest more in its logs:
- it's a centralized place
- the footprint is reduced/narrowed in web

That said, I understand it is a broader topic that we will eventually address, so let's leave it out from here 😅

arbulu89 (commented):
Hey @antgamdia ,
Have you tested this with a real failure?
The body of the failure will be a json, so we could consider doing some prettying or similar.
Just a suggestion to at least give a go.

PD:
I'm with @nelsonkopliku in the fact that we are duplicating the same error in `web` and `agent` logs but since some user asked for this...

arbulu89:
> Hey @antgamdia , Have you tested this with a real failure? The body of the failure will be a json, so we could consider doing some prettying or similar. Just a suggestion to at least give a go.
> 
> PD: I'm with @nelsonkopliku in the fact that we are duplicating the same error in `web` and `agent` logs but since some user asked for this...

It looks like it already does a good json output. Example (with modified `web` to create an error):

```
ERROR Error while running discovery discovery=host_discovery error=something wrong happened while publishing data to the collector. Status: 422, Agent: fb2c6b8a-9915-5969-a6b7-8b5a42de1971, discovery: host_discovery, body: {"errors":[{"title":"Invalid value","source":{"pointer":"/a"},"detail":"can't be blank"}]}

```

@abravosuse this feature was asked by some customer/premium engineer, right?

arbulu89 (approved):
good!
I think we should check `err` returned by `io.ReadAll(resp.Body)`. It is a normal golang practice

antgamdia:
> Have you tested this with a real failure?

Yes, I added this change exactly for that reason: to debug my changes in trento-web when dealing with https://github.com/trento-project/agent/pull/610. It was useful, as I just used an existing trento instance I had, but I did not want to dig into the trento-web logs, I just wanted early feedback on the call site.

> the output might be heavier and so the footprint on filesystem

Yes, I agree... but not being able to send the logs seems like a legit case for being verbose. Imagine issues with proxies where requests are not even able to reach trento-web. Having the response in the logs can be really useful for troubleshooting.

I'll be addressing the PR feedback and comments shortly. Thanks!

Edit: done https://github.com/trento-project/agent/pull/623/commits/2ca954351f8445f4be01b18ea22c1a4cb67c222f

abravosuse:
> @abravosuse this feature was asked by some customer/premium engineer, right?

@arbulu89 yes, it was one request -among a few others- made by Rafael Montes on June 25th via Slack.
@antgamdia when this is implemented, I would like to formally let Rafael know.

PR #633, opened 2026-07-21
Regenerate `_generated_wsdl` from SLES 16.1

antgamdia:
### Description

This pull request updates the generated (using the [instructions](https://github.com/trento-project/agent/blob/b11be332acd4bb0affcf2be64d135697be5eadf5/internal/core/sapsystem/sapcontrolapi/README.md) )SAPControl API client code, mainly to add support for setting an ABAP server as inactive or active, similar to what can be done in the SAP SM51 transaction. The changes also include a package rename for consistency.

* Added the `ABAPSetServerInactive` request and `ABAPSetServerInactiveResponse` response types to allow toggling the ABAP server's active/inactive state. [[1]](diffhunk://#diff-99190fe76380da299c8392c72d7432ae5aebc607289c988ed5f3983e066d711bR674-R679) [[2]](diffhunk://#diff-99190fe76380da299c8392c72d7432ae5aebc607289c988ed5f3983e066d711bR2661-R2663)
* Extended the `SAPControlPortType` interface with `ABAPSetServerInactive` and `ABAPSetServerInactiveContext` methods, enabling both synchronous and context-aware calls for setting server state.
* Implemented the corresponding methods on the `sAPControlPortType` struct to handle the new operations.
* Renamed the package from `sapcontrol` to `sapcontrolapi` to better reflect its purpose and align with project structure.

Related [TRNT-4574](https://jira.suse.com/browse/TRNT-4574)

### How was this tested?

UT

### Documentation changes

No

### Additional information

This file is not being used (or even compiled) at all. It is the source from which we hand-craft another file: https://github.com/trento-project/agent/blob/b11be332acd4bb0affcf2be64d135697be5eadf5/internal/core/sapsystem/sapcontrolapi/webservice.go


nelsonkopliku (commented):
Looks good, however question: are we sure this is backward compatible with sles 15? 👀

antgamdia:
Thanks!

> are we sure this is backward compatible with sles 15? 👀

That's not really important in this case. This file is not being used (or even compiled) at all. It is the source from which we hand-craft another file: https://github.com/trento-project/agent/blob/b11be332acd4bb0affcf2be64d135697be5eadf5/internal/core/sapsystem/sapcontrolapi/webservice.go

I followed: https://github.com/trento-project/agent/blob/b11be332acd4bb0affcf2be64d135697be5eadf5/internal/core/sapsystem/sapcontrolapi/README.md


Since the new stuff is neither relevant nor used elsewhere, I did not add it to that manually created file, so the changes are harmless.

nelsonkopliku (approved):

PR #643, opened 2026-07-23
Rename go module to `agent/v3`

antgamdia:
### Description

It turns out that the changes in #638 are not enough. Our go module is being rejected by pkg.go.dev, because it is not compliant with the naming policies. If our version is v3.y.z, then the go module has to be named `agent/v3`.
This PR is simply to rename from `agent` to `agent/v3`. No side effects are expected, as this package is not being consumed in other internal projects.

Related #[TRNT-4386](https://jira.suse.com/browse/TRNT-3886)

### How was this tested?

CI

### Documentation changes

No

### Additional information

See how https://sum.golang.org/lookup/github.com/trento-project/agent@v3.1.2 yields:

````console
not found: github.com/trento-project/agent@v3.1.2: invalid version: module contains a go.mod file, so module path must match major version ("github.com/trento-project/agent/v3")
````

### CODE REVIEW

File: internal/core/hosts/discovered_host.go [hunk 1]
```diff
@@ -5,7 +5,7 @@ package hosts
 
 import "time"
 
-import "github.com/trento-project/agent/internal/core/systemd"
+import "github.com/trento-project/agent/v3/internal/core/systemd"
```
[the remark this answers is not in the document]
antgamdia: Linting issues are handled in other PRs, thanks anyway!

Makefile
```diff
@@ -3,8 +3,8 @@
 
 VERSION ?= $(shell ./hack/get_version_from_git.sh)
 INSTALLATIONSOURCE ?= "Community"
-LDFLAGS = -X github.com/trento-project/agent/version.Version="$(VERSION)"
-LDFLAGS := $(LDFLAGS) -X github.com/trento-project/agent/version.InstallationSource="$(INSTALLATIONSOURCE)"
+LDFLAGS = -X github.com/trento-project/agent/v3/version.Version="$(VERSION)"
```
arbulu89, line 6: question:
does this path makes sense? Does it even exist?

arbulu89 (approved):
I don't know what to say.
For me, it looks simply more burden and maintenance work for a documentation page so far nobody really uses.
But well, so be it...

Makefile
antgamdia: I think so, it is the new name of the module, so for the linker to be able to change it, we need to use this path. See:

```bash
go build -ldflags "-X github.com/trento-project/agent/v3/version.Version=dummy-1.2.3 -X github.com/trento-project/agent/v3/version.InstallationSource=Community" -o /tmp/trento-agent-dummy .
````

```bash
/tmp/trento-agent-dummy version

Trento installed from Community version dummy-1.2.3
built with go1.25.12 linux/amd64
````

antgamdia:
> For me, it looks simply more burden and maintenance work for a documentation page so far nobody really uses.

Not really, if we want to consume this project from another go project, it becomes really tedious if the module name does not align. It is not just the pkg documentation, but the ability for others to consume the project.
In the past, I had to import and wasn't able to do it easily...  😅

antgamdia:
More issues... https://proxy.golang.org/github.com/trento-project/agent/v3/@v/v3.1.3-0.20260729105833-672083e58df0.info 😅

`not found: create zip: version/version.go: case-insensitive file name collision: "VERSION" and "version"
`
  I thought it only happens to me (in my case-insensitive fs), but it makes go pkg complain... I'm gonna do the same approach we do in mcp server for the versions in a follow-up PR.

PR #644, opened 2026-07-29
Move version to internal/version

antgamdia:
### Description

Following up on https://github.com/trento-project/agent/pull/643#issuecomment-5116933479, this PR moves the `version.go` to a sub-dir, so that it does not collide with the `VERSION` file in non-case-sensitive filesystems.

Besides, it adds some default values and avoids using global variables to store that information (like we do [in other projects](https://github.com/trento-project/mcp-server/blob/main/cmd/root.go#L261C1-L269C2))

Related #[TRNT-4386](https://jira.suse.com/browse/TRNT-3886)

### How was this tested?

IRL

### Documentation changes

 No

### Additional information

Example default output:

```bash
> go run github.com/trento-project/agent/v3 version
Trento installed from devel version devel
```


arbulu89 (approved):

antgamdia:
It worked!

https://pkg.go.dev/github.com/trento-project/agent/v3

PR #645, opened 2026-07-29
Fix saptune parsing with output containing colors

antgamdia:
### Description

This pull request refactors how the `saptune` command is executed to improve reliability when parsing JSON output and enhance logging for debugging. The main change is switching from capturing both stdout and stderr to capturing only stdout, which prevents non-JSON output (such as warnings on stderr) from breaking JSON parsing. All related tests are updated to reflect this change in the executor method.

* Updated `runSaptuneJSON` in `saptune.go` to use `OutputContext` instead of `CombinedOutputContext`, ensuring only stdout is captured for JSON parsing and adding more detailed logging for command execution and errors.
* Updated all test cases in `saptune_test.go` to expect `OutputContext` calls instead of `CombinedOutputContext`, keeping tests consistent with the new implementation. 

This is to fix this error:

```console
Jul 28 09:00:05 hana-s2-db1 trento-agent[2670]: 2026-07-28 09:00:05 ERROR fact gathering error: saptune-cmd-error - error executing saptune command: invalid character '\x1b' looking for beginning of value 
```

Fixes #[TRNT-4595](https://jira.suse.com/browse/TRNT-4595)

### How was this tested?

IRL

### Documentation changes

No

### DIFF
--- a/internal/core/saptune/saptune.go
+++ b/internal/core/saptune/saptune.go
@@ -159,7 +159,21 @@ func (s *saptuneClient) runSaptune(ctx context.Context, args ...string) ([]byte,
 }
 
 func (s *saptuneClient) runSaptuneJSON(ctx context.Context, args ...string) ([]byte, error) {
+	slog.Info("Running saptune command with json format", "args", args)
+
 	prependedArgs := append([]string{"--format", "json"}, args...)
 
-	return s.runSaptune(ctx, prependedArgs...)
+	// Using OutputContext instead of CombinedOutputContext to avoid capturing stderr,
+	// which may contain non-JSON output (e.g., warnings) that would break JSON parsing.
+	output, err := s.executor.OutputContext(ctx, "saptune", prependedArgs...)
+	if err != nil {
+		slog.Error("error executing saptune command", "args", prependedArgs, "error", err)
+
+		return output, fmt.Errorf("error executing saptune command: %w", err)
+	}
+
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
Hey @antgamdia ,
Before even reviewing, I want to understand why `saptune` is returning color characters. Can this be a failure in saptune part? Are we hiding something that should be fixed somewhere else?
Trento shouldn't run commands that return colored output. I think all commands we use have this in common, or an option to remove colors. (For example, `saptune` has `--force-color` option).
With this change, we are not using the `raw` output anymore, it is modified, and it worries me a bit.
Usually, all comands with colored output are always removed from piped results (which is what happens with golang, we shouldn't have this "terminal" helper chars there).
Maybe in this case, as it is a json output, saptune doesn't remove the chars from there, but maybe it should...

antgamdia:
As we discussed offline, for now, let's discard the stderr in the saptune command (we are not doing anything useful with it anyway), and defer the ANSI stripping thing if future issues arise.

In parallel, the saptune team will remove the ANSI escape sequences (colors mainly) from the error output if the json format is requested.

arbulu89 (approved):
Thank you @antgamdia ,
I think this is the perfect solution.
@abravosuse once this is merged and rolling package created, the issue you have with the check should be fixed.
Marking the fix for 3.1.3 patch release

